PR #48, opened 2020-01-15
MOON-41 : Allow SecondaryNav to be hidden

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-41

## Description

 Allow SecondaryNav to be visible or hidden

## Tests

- [X] Unit Test


tdraier:
> So I keep the state of the component to manage the visibility because it's the default behavior of the component.
> 
> * I rename `isVisible` by `isDefaultVisible`
> * I rename `onToggleVisible` by `onToggle`
> 
> What do you think?

It looks quite minor, but I would prefer to have `onToggled` instead of `onToggle` - `onToggle` let me think that the component requests to be toggled (and so, be controlled), whereas `onToggled` would be more clearly implies that the component did the toggle, and just notifies you it's done. And again, if we handle animation , it would be great to have `onToggling` called before the animation and `onToggled` after. But for now `onToggled` can be enough (but should be called after the animation ended)

src/components/ResizableBox/ResizableBox.jsx
```diff
@@ -65,23 +67,31 @@ ResizableBox.propTypes = {
...
      */
     defaultSize: PropTypes.shape({
         width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
         height: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
     }),
 
+    /**
+     * Manage the size
+     */
+    size: PropTypes.shape({
```
tdraier: If size is exposed we should also expose onResize* callback to control resizing.

PR #55, opened 2020-02-02
MOON-55: remove undefined height:100%

tdraier:
## JIRA

https://jira.jahia.org/browse/MOON-55

## Description

2 main changes : 
- removed all usage of 100% height, which in most of the case does not mean anything and give inconsistent behaviour. Since the layout relies on flexbox, use flex attributes.
- do not set layout information on the top element of a component, this is reponsability of the container to define how it will be sized. Especially , flexFluid class , or % sizes , do not make sense on the top element if you don't know how the container will be layouted.
- add className everywhere to be able to set an additional class on the top element of the component

## Tests

- [ ] Unit Test skeleton

## Checklist

- [ ] Required props and default values
- [ ] Example in storybook

## Documentation

- [ ] README documentation


src/components/SecondaryNav/SecondaryNav.stories.jsx
```diff
@@ -14,13 +14,11 @@ storiesOf('Components|SecondaryNav', module)
...
-            </SecondaryNav>
-        </div>
+        <SecondaryNav
+            header={text('Title', 'my header')}
+        >
+            My content here
+        </SecondaryNav>
     ))
     .add('with header images', () => (
         <div style={{transform: 'scale(1)', height: '100vh'}}>
```
ffffffelix: We should remove this `div` too

PR #56, opened 2020-02-03
MOON-12: Create dropdown component

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-12

## Description

- Create the dropdown component
- Create the menu component
- Create the ListItem component

## Tests

- [ ] Unit Test skeleton

## Checklist

- [X] All files present ( component - md - scss - spec - stories )
- [X] Required props and default values
- [X] Example in storybook

## Documentation

- [ ] README documentation


src/components/Menu/Menu.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Coordinate to position the menu
+     */
+    position: PropTypes.shape({
+        top: PropTypes.string,
+        right: PropTypes.string,
+        bottom: PropTypes.string,
+        left: PropTypes.string
+    }),
```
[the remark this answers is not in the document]
tdraier: Material-ui defines a complex set of props for that (4 props), but that's quite clean and cover the cases you describe :
- `anchorReference` prop is either 'anchorEl' (relative to an element) or 'anchorPosition' (absolute)
- if set to 'anchorEl', you need to set `anchorEl` prop to the element , and `anchorOrigin` prop to set the relative position to this element : `{ horizontal: number | 'left'| 'center' | 'right', vertical: number | 'top' | 'center' | 'bottom' }`
- if set to 'anchorPosition', you need to set `anchorPosition` prop with the absolute position : `{ left: number, top: number }`
https://material-ui.com/api/popover/

src/components/Menu/Menu.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Content of the dropdown
+     */
+    children: PropTypes.node.isRequired,
+
+    /**
+     * Menu is display
+     */
+    isDisplay: PropTypes.bool.isRequired,
```
tdraier: `isDisplayed`

src/components/Menu/Menu.jsx
```diff
@@ -0,0 +1,53 @@
...
+    return null;
+};
+
+Menu.defaultProps = {
+    maxWidth: 'fit-content',
+    position: {top: '0', left: '0'}
+};
+
+Menu.propTypes = {
+    /**
```
tdraier: We will need `onMouseEnter()` , `onMouseLeave()` , and ideally `onClosed()` if there's a fade out, to be called after the transition

src/components/Menu/Menu.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'clsx';
+import styles from './Menu.scss';
+
+export const Menu = ({children, isDisplay, maxWidth, position, className}) => {
```
tdraier: Shouldn't we handle the backdrop overlay in Menu instead of DropDown ?

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,270 @@
...
+        setIsOpened(true);
+    };
+
+    const handleSelect = e => {
+        const selectOption = e.target.textContent;
+        const options = isGrouped ? data.map(items => items.options).flat() : data;
+        const newItem = options.filter(option => option.label === selectOption)[0];
+
+        if (!newItem.isDisabled) {
+            setCurrentOption(newItem);
```
tdraier: Do not use state, call onChange if it's different from current value

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,270 @@
...
+    isDisabled,
+    maxWidth,
+    variant,
+    size,
+    icon,
+    onChange,
+    className
+}) => {
+    const [isOpened, setIsOpened] = useState(false);
+    const [currentOption, setCurrentOption] = useState(defaultOption ? defaultOption : {label: placeholder, value: null});
```
tdraier: Delete state, makes it controlled by default

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,270 @@
...
+                    ...PropTypesOptions
+                })
+            )
+        })
+    ).isRequired,
+
+    /**
+     * Option selected by default
+     */
+    defaultOption: PropTypes.shape({...PropTypesOptions}),
```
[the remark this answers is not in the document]
tdraier: And actually , "value" instead of "defaultValue" if controlled

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,257 @@
...
+        setAnchorEl(e.currentTarget);
+        setIsOpened(true);
+    };
+
+    const handleSelect = (e, item) => {
+        if (!item.isDisabled && item.value !== value) {
+            onChange(e, item);
+        }
+
+        if (!item.isDisabled) {
```
tdraier: I think it would be better to use the return of  onChange() to see if we should close or not the dropdown (so just a simple boolean, true if ok to close and false otherwise). If the controller decides that the selection is invalid, it could be useful to keep the drop down open. But in that case we should call onChange even if `item !=== value`

src/components/ListItem/ListItem.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const handleClick = e => {
+        onClick(e);
+    };
+
+    const handleMouseEnter = e => {
+        onMouseEnter(e);
+    };
+
+    const handleMouseLeave = e => {
```
tdraier: No need to redefine the handlers here, you can directly use onClick() / onMouseEnter() / onMouseLeave()

src/components/Menu/Menu.jsx
```diff
@@ -0,0 +1,147 @@
...
+    if (maxWidth) {
+        styleMenu.maxWidth = maxWidth;
+    }
+
+    // ---
+    // Render
+    // ---
+    if (isDisplayed) {
+        return (
+            <div className={classnames(styles.menu_wrapper)}>
```
tdraier: We will need the backdrop to be optional to be able to use the menu as a submenu (only the root menu has a backdrop)

src/components/Menu/Menu.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+    const setPositioningStyles = useCallback(() => {
+        const positioning = getPositioningStyle(anchorEl);
+        const positioningStyles = {
+            top: `calc(${positioning.top}px + ${anchorPosition.top})`,
+            left: `calc(${positioning.left}px + ${anchorPosition.left})`
+        };
+
+        setStylePosition(positioningStyles);
+    }, [getPositioningStyle, anchorEl, anchorPosition]);
```
tdraier: I'm not very familiar with useCallback, but what does it mean to have another callback as a dependency ? Shouldn't we use a simple useMemo ( `const positionStyle = useMemo(() => ...), [anchorEl])` , then `setPositioningStyles = useCallback(()=> {...}, [positionStyle, anchorPosition]` ? Or even more simple, if getPositioningStyle() is only used in setPositioningStyles(), just use one single useCallback().

ffffffelix: I have to confess this part of code is mostly inspired by the component popover from materia-ui.

src/components/Dropdown/Dropdown.jsx
ffffffelix: But it’s not weird to trigger the `onChange` even if the value doesn’t change?

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,257 @@
...
+    // ---
+    const handleOpenMenu = e => {
+        setMinWith(`${e.currentTarget.offsetWidth}px`);
+        setAnchorEl(e.currentTarget);
+        setIsOpened(true);
+    };
+
+    const handleSelect = (e, item) => {
+        if (!item.isDisabled && item.value !== value) {
+            onChange(e, item);
```
tdraier:
```suggestion
            canClose = onChange(e, item);
```

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,257 @@
...
+    // Functions to handle events
+    // ---
+    const handleOpenMenu = e => {
+        setMinWith(`${e.currentTarget.offsetWidth}px`);
+        setAnchorEl(e.currentTarget);
+        setIsOpened(true);
+    };
+
+    const handleSelect = (e, item) => {
+        if (!item.isDisabled && item.value !== value) {
```
tdraier:
```suggestion
        let canClose = !item.isDisabled;
        if (!item.isDisabled && item.value !== value) {
```

src/components/Dropdown/Dropdown.jsx
tdraier:
```suggestion
        if (canClose) {
```

PR #59, opened 2020-02-11
Moon 58: Improve accordion - makes it controllable

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-58

## Description

- Authorize the opening/closing of accordion by programmation
- UI update

## Breaking change

⚠️ The props `openByDefault ` has been renamed `defaultOpenedItem`

## Tests

- [ ] Unit Test skeleton

## Checklist

- [X] Required props and default values
- [X] Example in storybook

## Documentation

- [ ] README documentation

### DIFF
--- a/src/components/Accordion/Accordion.jsx
+++ b/src/components/Accordion/Accordion.jsx
@@ -37,12 +37,12 @@ Accordion.propTypes = {
     ]).isRequired,
 
     /**
-     * AccordionItem's id opened by default (isMultipleOpenable ? Array : String)
+     * AccordionItem's id opened by default
      */
     defaultOpenedItem: PropTypes.string,
 
     /**
-     * AccordionItem's id open (isMultipleOpenable ? Array : String)
+     * AccordionItem's id open
      */
     openedItem: PropTypes.string,
 
--- a/src/components/Accordion/Accordion.jsx
+++ b/src/components/Accordion/Accordion.jsx
@@ -1,21 +1,55 @@
-import React, {useState} from 'react';
+import React, {useState, useEffect} from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'clsx';
 import {AccordionContext} from './Accordion.context';
 import {AccordionItem} from './AccordionItem';
 import styles from './Accordion.scss';
 
-export const Accordion = ({children, openByDefault, isReversed, className, ...props}) => {
-    const [currentItem, setCurrentItem] = useState(openByDefault ? openByDefault : null);
-    const [reversed] = useState(isReversed);
+export const Accordion = ({children, defaultOpenedItemId, openedItemId, isReversed, className, isMultipleOpenable, ...props}) => {
+    const [stateOpenedItemId, setStateOpenedItemId] = useState([]);
 
-    function defineCurrentItem(id) {
-        setCurrentItem(prevState => (prevState === id ? null : id));
-    }
+    const setOpenedItemId = id => {
+        console.log('call setOpenedItemId');
+        setStateOpenedItemId(prevState => {
+            console.log(prevState);
+            console.log('---');
+            if (prevState === null) {
+                return [id];
+            }
+
+            if (isMultipleOpenable) {
+                return prevState.includes(id) ? prevState.filter(i => id !== i) : [...prevState, id];
+            }
+
+            return prevState.includes(id) || id === null ? [] : [id];
+        });
+    };
+
+    useEffect(() => {
+        console.log('call useEffect');
+        console.log(defaultOpenedItemId);
+        console.log(openedItemId);
+        setStateOpenedItemId(defaultOpenedItemId && !openedItemId ? defaultOpenedItemId : openedItemId);
+    }, [defaultOpenedItemId, openedItemId]);
+
+    const provider = {
+        currentItemId: stateOpenedItemId,
+        setOpenedItemId,
+        isReversed: isReversed
+    };
 
     return (
-        <AccordionContext.Provider value={{currentItem, defineCurrentItem, reversed}}>
-            <div {...props} className={classnames(className, 'flexFluid', styles.accordion, isReversed ? styles.accordion_reversed : null)}>
+        <AccordionContext.Provider value={provider}>
+            <div className={
+                    classnames(
+                        className,
+                        'flexFluid',
+                        styles.accordion,
+                        isReversed ? styles.accordion_reversed : null
+                    )
+                }
+                 {...props}
+            >
                 {children}
             </div>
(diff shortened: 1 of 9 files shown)


src/components/Accordion/Accordion.jsx
```diff
@@ -1,29 +1,74 @@
...
 import classnames from 'clsx';
 import {AccordionContext} from './Accordion.context';
 import {AccordionItem} from './AccordionItem';
 import styles from './Accordion.scss';
 
-export const Accordion = ({children, openByDefault, isReversed, className, ...props}) => {
-    const [currentItem, setCurrentItem] = useState(openByDefault ? openByDefault : null);
-    const [reversed] = useState(isReversed);
+export const Accordion = ({children, defaultOpenedItemId, openedItemId, isReversed, className, isMultipleOpenable, ...props}) => {
+    const [stateOpenedItemId, setStateOpenedItemId] = useState(defaultOpenedItemId);
```
tdraier: If the component is controlled, it should not have a state. Only way to make this clean and stable with state hooks is to have separate components for controlled and uncontrolled versions

src/components/Accordion/Accordion.jsx
```diff
@@ -1,29 +1,74 @@
...
+            }
+
+            return prevState === id ? null : id;
+        });
+    };
+
+    useEffect(() => {
+        // Update the state when the prop openedItemId has changed
+        // Because of the first render with have to check if defaultOpenedItemId is set
+        setStateOpenedItemId(defaultOpenedItemId && !openedItemId ? defaultOpenedItemId : openedItemId);
```
tdraier: Setting the state from a prop, unless it's a default value, is an anti-pattern

src/components/Accordion/Accordion.jsx
```diff
@@ -47,9 +92,19 @@ Accordion.propTypes = {
...
     /**
-     * AccordionItem's id to set an item open by default
+     * AccordionItem's id opened by default (isMultipleOpenable ? Array : String)
+     */
+    defaultOpenedItemId: idPropType,
+
+    /**
+     * AccordionItem's id open (isMultipleOpenable ? Array : String)
+     */
+    openedItemId: idPropType,
```
tdraier: Should also expose setOpenedItemId to control the component, if openedItemId is set

src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -7,29 +7,55 @@ import {AccordionContext} from '~/components/Accordion/Accordion.context';
...
+    }
 
     const handleClick = e => {
         if (!open) {
             onClickToOpen(e);
         }
 
         onClick(e);
-        context.defineCurrentItem(id);
+        context.setOpenedItemId(id);
```
tdraier: This one shoudl only call the setOpenedItemId prop from accordion - not modifying any state, if component is controlled

src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -7,29 +7,55 @@ import {AccordionContext} from '~/components/Accordion/Accordion.context';
...
+
+    if (context.isMultipleOpenable) {
+        open = context.currentItem ? context.currentItem.includes(id) : false;
+    } else {
+        open = context.currentItem === id;
+    }
 
     const handleClick = e => {
         if (!open) {
             onClickToOpen(e);
```
tdraier: Do you think we really need all these callbacks ? onClickToOpen / onClickToClose / onClick, in addition to the setOpenedItem of Accordion ?

ffffffelix: Maybe not, we just keep `onClick`? Or I remove all callbacks?

ffffffelix: When we use the UncontrolledAccordion we haven't access to `setOpenedItem`

src/components/Accordion/ControlledAccordion.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Additional classname
+     */
+    className: PropTypes.string,
+
+    /**
+     * Function to set accoridonItem opened
+     */
+    setOpenedItem: PropTypes.func
```
tdraier: To be consistent with other callback naming, maybe we could name this onOpenItem ( and onCloseItem , like in TreeView ) ?

src/components/Accordion/UncontrolledAccordion.jsx
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * AccordionItem's id opened by default
+     */
+    defaultOpenedItem: PropTypes.string,
+
+    /**
+     * Additional classname
+     */
+    className: PropTypes.string
+};
```
tdraier: We may need callbacks in uncontrolled to tell when an accordion has been selected - onItemOpened / onItemClosed ?

src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -15,21 +15,41 @@ export const AccordionItem = ({id, label, icon, onClickToOpen, onClick, onClickT
         }
 
         onClick(e);
```
tdraier: If we want to keep callback here, one is enough - onClickToOpen and onClickToClose are superfluous ( but pass open as a parameter to onClick )

src/components/Accordion/ControlledAccordion.jsx
ffffffelix: In `TreeView` we don’t have access to the item while for the `Accordion`we have access to `AccordionItem`.

`setOpenedItem` is used from the `AccordionContext` to set the opened item. It is not superfluous and add complexity to split it to `onOpenItem` and `onCloseItem` to do almost the same thing?

src/components/Accordion/UncontrolledAccordion.jsx
ffffffelix: it doesn't duplicate with the callback on the `onClick` event on `AccordionItem`?
Should we manage the open/close callback from `Accordion` or `AccordionItem`?

src/components/Accordion/ControlledAccordion.jsx
ffffffelix: As discussed with Thomas we name it `onSetOpenedItem`

src/components/Accordion/AccordionItem/AccordionItem.jsx
ffffffelix: We just have `onClick` with `open` as second parameters

PR #66, opened 2020-02-19
MOON-30: Add github workflow to deploy storybook on github pages

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-30

## Description

Deploy storybook to github pages when moonstone is released

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -1,0 +1,26 @@
+name: CI Deploy storybook
+
+on:
+  release:
+    types: [published, unpublished]
+
+jobs:
+  build-storybook:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,26 @@
+name: CI Deploy storybook
+
+on:
+  release:
+    types: [published, unpublished]
```
tdraier: why unpublished ?

ffffffelix: To have storybook sync with the current version

tdraier: But even if you unrelease, it will just regenerate storybook from the master , so what's the difference ? (actually it would even be a more recent version than the previously released one ?)

ffffffelix: You right! I removed the `unpublished` trigger

PR #69, opened 2020-02-20
MOON-67 : Fixed styling, handle root node

tdraier:
## JIRA

https://jira.jahia.org/browse/MOON-XXX

## Description

XXX component definition

## Tests

- [x] Unit Test skeleton

## Checklist

- [x] Example in storybook

## Documentation

- [x] README documentation

### DIFF
--- a/src/components/TreeView/ControlledTreeView.jsx
+++ b/src/components/TreeView/ControlledTreeView.jsx
@@ -7,11 +7,17 @@ import {Typography} from '~/components';
 import PropTypes from 'prop-types';
 import React from 'react';
 
+const imgSizes = {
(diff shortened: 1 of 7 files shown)


src/components/TreeView/ControlledTreeView.jsx
```diff
@@ -7,11 +7,17 @@ import {Typography} from '~/components';
 import PropTypes from 'prop-types';
 import React from 'react';
 
+const imgSizes = {
+    small: '12px',
+    default: '16px'
```
ffffffelix: Maybe we should use a global CSS variable to define icon sizes?

PR #78, opened 2020-02-21
MOON-51 implement buttons, update API and sass, implement storybook, implement test, make all icons available in all tests

DameniMilo:
## JIRA
https://jira.jahia.org/browse/MOON-51

## Description
- implement buttons
- update API and sass
- implement storybook
- implement test
- make all icons available in all tests

## Tests

- [x] Unit Test skeleton

## Checklist
- [x] Example in storybook

## Documentation
- [ ] README documentation

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -4,9 +4,9 @@ import styles from './Button.scss';
 import classnames from 'clsx';
 import {Typography} from '../Typography';
 
-export const ButtonSizes = ['small', 'default', 'big'];
-export const ButtonVariants = ['default', 'ghost', 'outlined'];
-export const ButtonColors = ['default', 'accent', 'success', 'warning', 'danger'];
+export const buttonSizes = ['small', 'default', 'big'];
+export const buttonVariants = ['default', 'ghost', 'outlined'];
+export const buttonColors = ['default', 'accent', 'success', 'warning', 'danger'];
 
 export const Button = ({label, onClick, size, isReversed, isDisabled, icon, variant, color, className, ...props}) => {
     return (
@@ -57,7 +57,7 @@ Button.propTypes = {
     /**
      * Icon size
      */
-    size: PropTypes.oneOf(ButtonSizes),
+    size: PropTypes.oneOf(buttonSizes),
 
     /**
      * Icon name, if it's empty the button has no icon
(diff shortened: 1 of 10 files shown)


src/components/Button/Button.jsx
```diff
@@ -2,31 +2,50 @@ import React from 'react';
...
+            {icon && <icon.type {...icon.props} size={(size === 'big') ? 'default' : size}/>}
+            {label &&
+                <Typography isNowrap component="span" variant="button" isUpperCase={size === 'big'} weight={(size === 'small') ? 'light' : 'default'}>
+                    {label}
+                </Typography>}
         </button>
     );
 };
 
 Button.defaultProps = {
```
cedricmoitrier: What about having a default value for onClick ?

src/components/Button/Button.jsx
```diff
@@ -2,31 +2,50 @@ import React from 'react';
...
 import styles from './Button.scss';
 import classnames from 'clsx';
+import {Typography} from '../Typography';
 
-export const ButtonSizes = ['small', 'medium', 'big'];
-export const ButtonVariants = ['default', 'primary', 'ghost'];
-export const ButtonColors = ['accent', 'inherit', 'success', 'warning', 'danger', 'reverse'];
+export const ButtonSizes = ['small', 'default', 'big'];
+export const ButtonVariants = ['default', 'ghost', 'outlined'];
+export const ButtonColors = ['default', 'accent', 'success', 'warning', 'danger'];
```
cedricmoitrier: Shouldn't those be camelCased?

src/components/Button/Button.jsx
ffffffelix: As `onClick` is required, do really need a default value?

src/components/Button/Button.spec.js
```diff
@@ -1,43 +1,113 @@
...
+
+        const button2 = shallow(<Button variant="default" onClick={() => {}}/>);
+        expect(button2.querySelector('.variant_default').exists()).toBeTruthy();
+        expect(button2.querySelector('.variant_outlined').exists()).toBeFalsy();
+        expect(button2.querySelector('.variant_ghost').exists()).toBeFalsy();
+    });
+
+    it('should use the specified variant', () => {
+        // eslint-disable-next-line array-callback-return
+        ButtonVariants.map(variant => {
```
cedricmoitrier: You could/should use <array>.forEach(...) instead of <array>.map(...). You would not need to disable the eslint rule then.

src/components/Button/Button.jsx
cedricmoitrier: Would simplify the writing of unit test at least :-)

src/components/Button/Button.jsx
DameniMilo: probably

src/components/Button/Button.spec.js
DameniMilo: good point

src/components/Button/Button.jsx
DameniMilo: I don't know, I would say no

PR #98, opened 2020-03-03
Moon 82 Header for secondary nav

AKarmanov:
## JIRA

https://jira.jahia.org/browse/MOON-82

## Description

We need to be able to have consistency in our headers in secondary nav. We want to do minimal styling and be able to just pass a component which will be centered and displayed properly.


src/components/SecondaryNav/SecondaryNavHeader/SecondaryNavHeader.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'clsx';
+import styles from './SecondaryNavHeader.scss';
+
+const SecondaryNavHeader = ({children}) => {
+    return (
+        <div className={classnames('.flexCol_center', '.alignCenter', styles.secondaryNavHeader)}>
```
ffffffelix: - You don't need the dot before classes
- I think is better to give a row context in case we want to align icon and text, what do you think?
```suggestion
        <div className={classnames('flexRow_center', 'alignCenter', styles.secondaryNavHeader)}>
```

src/components/SecondaryNav/SecondaryNavHeader/SecondaryNavHeader.scss
```diff
@@ -0,0 +1,10 @@
+.secondaryNavHeader {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
```
ffffffelix: If you use `flexCol_center` and `alignCenter` you don't need that
```suggestion
```

src/components/SecondaryNav/SecondaryNavHeader/SecondaryNavHeader.scss
```diff
@@ -0,0 +1,10 @@
...
+    width: 100%;
+    min-height: 3em;
+
+    padding: 5px;
```
ffffffelix: We use a base 8 for spacing. So you can use `4px` or `8px`.
It's better if you use our spacing variables `var(--spacing-nano)` or `var(--spacing-small)`

src/components/SecondaryNav/SecondaryNavHeader/SecondaryNavHeader.scss
```diff
@@ -0,0 +1,10 @@
+.secondaryNavHeader {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    width: 100%;
+    min-height: 3em;
```
ffffffelix: I think we should add a min-height on `secondaryNav_header`

AKarmanov: I think that in that case what I put in the header will not be centered and aligned no?

AKarmanov: unless I do height: 100%, I don't know which is better

ffffffelix:
I think we can also directly display the `header` props on `SecondaryNav`.
We don't need the wrapper anymore.
```
 <header className={classnames(styles.secondaryNav_header, 'flexRow_center', 'alignCenter')}>
        {header}
</header>
```
So you can use this markup on your new component `SecondaryNavHeader` it will be more semantic than a `div`

src/components/SecondaryNav/SecondaryNavHeader/SecondaryNavHeader.scss
```diff
@@ -2,5 +2,5 @@
     width: 100%;
     min-height: 3em;
 
-    padding: 5px;
+    padding: 4px;
```
ffffffelix:
```suggestion
    padding: var(--spacing-nano);
```

PR #99, opened 2020-03-03
Moon 36: Add active state to MenuItem

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-36

## Description
- Remove all states from `ListItem`, now all states are managed from `MenuItem`
- Add active state on `MenuItem`
- Allow opening the dropdown by pressing Enter when it's focused

## Tests

- [ ] Unit Test skeleton

## Checklist

- [ ] Required props and default values
- [ ] Example in storybook

## Documentation

- [X] README documentation

### DIFF
--- a/src/components/Dropdown/Dropdown.md
+++ b/src/components/Dropdown/Dropdown.md
@@ -9,8 +9,3 @@ Allow user to validate a value by pressing `Enter`
 To hide the menu click on an item of the list or click outside to it
 The selected item has a specific design
 Items can be grouped with a title
-
-
-## Implementation
-
-To allow to use the dropdown with keyboard don't forget to add `e.stopPropagation();` on the onChange callback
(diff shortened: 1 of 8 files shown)


src/components/Dropdown/Dropdown.md
```diff
@@ -9,3 +9,8 @@ Allow user to validate a value by pressing `Enter`
 To hide the menu click on an item of the list or click outside to it
 The selected item has a specific design
 Items can be grouped with a title
+
+
+## Implementation
+
+To allow to use the dropdown with keyboard don't forget to add `e.stopPropagation();` on the onChange callback
```
tdraier: Shouldn't it be done directly in DropDown component ?

ffffffelix: Yes, it should be better, but I don't find the right place to add it without break everything

ffffffelix: I find a way 🤘

ffffffelix:
> And so we keep the nice colors ? :-)

I use colors from Figma

PR #102, opened 2020-03-04
MOON-61: Add disabled state for ButtonGroup

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-61

## Description

Add disabled state for ButtonGroup

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.jsx
+++ b/src/components/ButtonGroup/ButtonGroup.jsx
@@ -4,7 +4,7 @@ import styles from './ButtonGroup.scss';
 import classnames from 'clsx';
 import {buttonSizes, buttonVariants, buttonColors} from '~/components/Button';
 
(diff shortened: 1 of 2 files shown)


src/components/ButtonGroup/ButtonGroup.jsx
```diff
@@ -27,6 +27,7 @@ export const ButtonGroup = ({size, isReversed, variant, color, className, childr
                             size={size}
                             variant={variant}
                             isReversed={isReversed}
+                            isDisabled={isDisabled}
```
tdraier: Si isDisabled overrides the button isDisabled ? I have cases where I need some buttons to be disabled , and some to be enabled.

ffffffelix: So I close this PR and we will manage the disabled state in each button

PR #107, opened 2020-03-06
MOON-90 add tab component

Taybou:


src/components/Tab/Tab.jsx
```diff
@@ -0,0 +1,119 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './Tab.scss';
+import classnames from 'clsx';
+import {Typography} from '../Typography';
+
+export const tabSizes = ['small', 'default', 'big'];
```
ffffffelix: As I can see the mockup just provide the medium size

src/components/Tab/Tab.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+export const Tab = ({label, onClick, size, isReversed, isDisabled, icon, variant, color, className, isSelected, ...props}) => {
+    let typoWeight = 'default';
+    if (size === 'small') {
+        typoWeight = 'light';
+    }
+
+    if (size === 'big') {
+        typoWeight = 'semiBold';
+    }
```
ffffffelix: There is just one size, you don't need that

src/components/Tab/Tab.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+    return (
+        <button
```
ffffffelix: Why you don't use the `Button` component?

FBerthelot: Add the possibility to override the main element. See  @vindhya if you want help on this.

src/components/Tab/Tab.jsx
```diff
@@ -0,0 +1,119 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './Tab.scss';
```
FBerthelot: I guess the name of this component is TabItem. The Tab component for me is the combinaisons of all TabItem ?

src/components/Tab/Tab.jsx
Taybou: what did you mean by override the main element?

I'll see with her :D 

Thanks

Taybou: I update to something more generic

src/components/Tab/Tab.jsx
Taybou: updated

src/components/Tab/Tab.jsx
Taybou: updated

src/components/Tab/Tab.scss
```diff
@@ -0,0 +1,8 @@
+.tab {
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    align-items: center;
```
ffffffelix: Use `flexRow_center` and `alignCenter` classes

PR #116, opened 2020-03-11
BACKLOG-12710 add property invisible to the Separator so that we can hide it it is the first-child, last-child or only-child

DameniMilo:
## JIRA
https://jira.jahia.org/browse/BACKLOG-12710

## Description
add property invisible to the Separator so that we can hide it it is the first-child, last-child or only-child

## Tests

- [x] Unit Test skeleton

## Checklist
- [x] Example in storybook

## Documentation
- [ ] README documentation

### DIFF
--- a/src/components/Separator/Separator.scss
+++ b/src/components/Separator/Separator.scss
@@ -1,6 +1,18 @@
 .separator {
     border: 0;
     background: var(--color-gray40);
+
+    &.invisible_first-child:first-child {
+        display: none;
+    }
+
+    &.invisible_last-child:last-child {
+        display: none;
+    }
+
+    &.invisible_only-child:only-child {
+        display: none;
+    }
 }
 
 .separator_horizontal {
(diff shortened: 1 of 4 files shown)


src/components/Separator/Separator.scss
```diff
@@ -1,6 +1,18 @@
 .separator {
     border: 0;
     background: var(--color-gray40);
+
```
tdraier: what if I want to have my separator to be invisible when there no item before OR after ?

DameniMilo: well I did not think of this one

tdraier: i think it should be the default behaviour :-)

DameniMilo: Humm nope, because we do have cases in moonstone that expect the separator to show when it's the only child or the first-child, so if we do that we will have to change the implementation of the primary nav :/

DameniMilo: But I'm adding a fourth option as we speak to handle your cases, not as the default one tho

tdraier: ok, so by default no invisible

DameniMilo: yes, by default always visible

PR #137, opened 2020-04-03
MOON-96: Fix styling issue on ButtonGroup hover

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-96

## Description

Fix styling issue on hovering a ButtonGroup with a default variant and an accent color (a double border was displayed).

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.scss
+++ b/src/components/ButtonGroup/ButtonGroup.scss
@@ -20,14 +20,6 @@
 
     // Manage separator
     .variant_default {
-        .color_default {
-            &:hover,
-            &:active,
-            &:focus {
-                border-left-color: currentColor !important;
-            }
-        }
-
         & + .variant_default {
             border-left-color: transparent;
         }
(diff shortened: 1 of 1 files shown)


src/components/ButtonGroup/ButtonGroup.scss
```diff
@@ -20,10 +20,12 @@
...
     .variant_default {
-        &:hover,
-        &:active,
-        &:focus {
-            border-left-color: currentColor !important;
+        .color_default {
+            &:hover,
+            &:active,
+            &:focus {
+                border-left-color: currentColor !important;
```
tdraier, line 27: Can't you add more specificity instead of using !important ? Maybe a &&:hover,&&:active,&&:focus ?

ffffffelix: Our stylelint config prevents this usage.

Anyway, the styling button has been updated here: [MOON-112: Design improvements](https://github.com/Jahia/moonstone/pull/133),  so we don't need this piece of code. I remove it

tdraier: stylelint prevents `&&` but allows `!important` ? a base rule would be in my opinion to just forbid any `!important` anywhere :-)

tdraier: would be good to add declaration-no-important to the config

PR #141, opened 2020-04-06
BACKLOG-13003 Collapse primary nav when an item is clicked

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13003

## Description

Collapse primary nav when an item is clicked

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.jsx
+++ b/src/components/PrimaryNav/PrimaryNav.jsx
@@ -63,14 +63,14 @@ NavHeader.propTypes = {
     headerLogo: PropTypes.node
 };
 
-export const PrimaryNav = ({headerLogo, top, bottom, headerCaption, modeIcon, isCollapsibleOnCLick, ...props}) => {
+export const PrimaryNav = ({headerLogo, top, bottom, headerCaption, modeIcon, isCollapsibleOnClick, ...props}) => {
     const [isExpanded, setExpanded] = useState(false);
 
     function toggleExpand() {
         setExpanded(!isExpanded);
     }
 
-    const collapseFunc = isCollapsibleOnCLick ? () => setExpanded(false) : () => {};
+    const collapseFunc = isCollapsibleOnClick ? () => setExpanded(false) : () => {};
 
     return (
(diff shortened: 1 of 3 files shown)


src/components/PrimaryNav/PrimaryNav.jsx
```diff
@@ -63,15 +63,17 @@ NavHeader.propTypes = {
     headerLogo: PropTypes.node
 };
 
-export const PrimaryNav = ({headerLogo, top, bottom, headerCaption, modeIcon, ...props}) => {
+export const PrimaryNav = ({headerLogo, top, bottom, headerCaption, modeIcon, isCollapsibleOnCLick, ...props}) => {
```
ffffffelix, line 66: Do we want to keep the choice collapsible on click or not?
I think we can force the collapse in all cases.

If we keep it, please fix the typo `isCollapsibleOnClick`

src/components/PrimaryNav/PrimaryNavItem/PrimaryNavItem.jsx
```diff
@@ -62,23 +73,35 @@ ItemTypeResolver.propTypes = {
...
+        <li
+            className={classnames(
+                styles.primaryNavItem,
+                {[styles.selected]: isSelected},
+                className
+            )}
+            title={label}
+            onClick={e => {
+                primaryNavContext.collapse();
+                if (onClick) {
```
ffffffelix, line 89: You can add this, to manage empty onClick props
```
PrimaryNavItem.defaultProps = {
    onClick: () => {}
}
```

src/components/PrimaryNav/PrimaryNav.jsx
AKarmanov: My original idea was to make this behaviour optional but seeing how contested it is and that there is no clear direction for what  should be done and given that not having it means that I won't have to modify any other module I decided to remove it :)

src/components/PrimaryNav/PrimaryNavItem/PrimaryNavItem.jsx
AKarmanov: good point

PR #147, opened 2020-04-10
MOON-116: Add animation on opening/closing accordion

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-116

## Description

- Add animation

### DIFF
--- a/src/components/Accordion/AccordionItem/AccordionItem.jsx
+++ b/src/components/Accordion/AccordionItem/AccordionItem.jsx
@@ -1,4 +1,4 @@
-import React, {useContext} from 'react';
+import React, {useContext, useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'clsx';
 import styles from './AccordionItem.scss';
@@ -7,6 +7,7 @@ import {AccordionContext} from '~/components/Accordion/Accordion.context';
 
 export const AccordionItem = ({id, label, icon, onClick, children, className, ...props}) => {
     const context = useContext(AccordionContext);
+    const [cssExpanded, setCssExpanded] = useState(null);
     const open = context.currentItem === id;
 
     const handleClick = (e, open) => {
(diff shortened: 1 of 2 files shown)


src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -7,28 +7,33 @@ import {AccordionContext} from '~/components/Accordion/Accordion.context';
...
+    const [cssExpanded, setCssExpanded] = useState(null);
     const open = context.currentItem === id;
 
     const handleClick = (e, open) => {
         onClick(e, !open);
         context.onSetOpenedItem(id);
     };
 
+    useEffect(() => {
+        setCssExpanded({[styles.accordionItem_opened]: open});
```
tdraier, line 19: I don't understand the need of a state here - why don't we use directly open ?

src/components/Accordion/AccordionItem/AccordionItem.scss
```diff
@@ -95,3 +98,18 @@ $accordionHeader_height: 48px;
         background-color: var(--color-dark60);
     }
 }
+
+// Animation
+.accordionItem_opened {
```
tdraier, line 103: please follow conventions for modifiers - .accordionItem.opened  (  .accordion { &.opened {} }

src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -7,28 +7,33 @@ import {AccordionContext} from '~/components/Accordion/Accordion.context';
 
 export const AccordionItem = ({id, label, icon, onClick, children, className, ...props}) => {
```
tdraier, line 8: Add callbacks for transition handling - onItemOpening / onItemOpened on Accordion before and after transition , also check what to do with onSetOpenedItem  (currently called before transition)

src/components/Accordion/AccordionItem/AccordionItem.jsx
ffffffelix: It's to trigger the animation after the render. We had some user feedback than they don't understand there is some accordion at the bottom of the page. We think to animate the opening of the accordion can improve this behavior.

tdraier: So it's only for the first render, if the accordion item is directly opened ? This has no impact when opening/closing the accordion after its first render ?

src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -7,28 +7,33 @@ import {AccordionContext} from '~/components/Accordion/Accordion.context';
...
     const handleClick = (e, open) => {
         onClick(e, !open);
         context.onSetOpenedItem(id);
     };
 
+    useEffect(() => {
+        setCssExpanded({[styles.accordionItem_opened]: open});
+    },
+    [open]);
+
```
tdraier, line 22: If only for the first render I would suggest something like that : 
```suggestion
    const [initialized, setInitialized] = useState(false);
    useEffect(() => {
        setInitialized(true)
    }, []);
    const isExpanded = open && initialized
```
And replace the cssExpanded  with just `{[styles.expanded]: isExpanded}`

PR #151, opened 2020-04-23
MOON-132: Fix TreeView alignement

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-132

## Description

Fix tree alignment between siblings elements with children with icon or not


vindhya:
Hi @ffffffelix, should this fix solve the alignment for when the sibling element does/doesn't have an icon? I.e., should the labels be aligned with each other or where the icon could be?

Maybe this screenshot will make it easier to understand what I'm asking:
image

PR #163, opened 2020-08-10
MOON-53: Build Changes

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-53

## Description

* Compiling scss to css
* Adds babel config file for build


src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -1,12 +1,12 @@
...
 import PropTypes from 'prop-types';
-import styles from './Breadcrumb.scss';
+import './Breadcrumb.scss';
 import classnames from 'clsx';
 import {ChevronRight} from '~/icons';
 
 export const Breadcrumb = ({children, className, ...props}) => {
     const classNames = classnames(
-        styles.breadcrumb,
+        'breadcrumb',
```
tdraier, line 9: where does this class comes from ?

src/components/PrimaryNav/PrimaryNav.scss
```diff
@@ -3,7 +3,7 @@
 $primaryNavExpanded: 300px;
 $primaryNav: 56px;
 
-.logoCaptionGroup {
+.moonstone-logoCaptionGroup {
```
tdraier, line 6: It's not directly related to this PR, but the class naming here is invalid, they are not prefix with the block name . All classes should start with moonstone-primaryNav_*

webpack.config.js
```diff
@@ -27,21 +27,22 @@ const ComponentsConfig = {
...
+                test: /\.css$/i,
+                exclude: /\.module\.scss$/i,
                 sideEffects: true,
                 use: [
                     'style-loader',
-                    // Translates CSS into CommonJS
+                    'css-loader',
                     {
-                        loader: 'css-loader',
+                        loader: 'sass-loader',
```
tdraier, line 37: I'm not quite sure about this, but why would we need a sass loader here to load css that are already compiled from sass ?

src/components/Accordion/AccordionItem/AccordionItem.jsx
```diff
@@ -18,19 +18,19 @@ export const AccordionItem = ({id, label, icon, onClick, children, className, ..
         <section
             {...props}
             className={classnames(
-                styles.accordionItem,
-                {[styles.reversed]: context.isReversed},
+                'moonstone-accordionItem',
+                {reversed: context.isReversed},
```
tdraier, line 22: shouldn't we prefix all classes, including modifiers classes ?

src/components/Breadcrumb/Breadcrumb.jsx
vindhya: Good question 🙂 I saw it in the original jsx file so I left it in but I've removed it now.

src/components/PrimaryNav/PrimaryNav.scss
vindhya: Ok! I've updated a bunch of class names on this component.

webpack.config.js
vindhya: I think it was left over from a previous change but I tested it out with just css-loader and everything seems to be fine 👍🏾

src/components/Accordion/AccordionItem/AccordionItem.jsx
vindhya: I didn't add the prefix to the modifier classes because they require the top level 'moonstone-*' to be present as well, so it's already specific enough. Although, do you think that reasoning would be confusing for other people creating components?

E.g., 'reversed' is always used with 'moonstone-accordionItem' or 'moonstone-accordionItem_header' (https://github.com/Jahia/moonstone/blob/scss-build/src/components/Accordion/AccordionItem/AccordionItem.scss#L19)

tdraier: I was only thinking of prefixing them, because if somebody is also defining classes with same names (like "reversed"), we may get some styles from here which are not wanted .. for the opposite I agree that somebody using "reversed" will not get moonstone styles since it require the top-level class. Still I'm not fond of prefixing all classes, but there's a risk.

vindhya: Ok! I will rename the all the modifier classes (and any other classes that aren't already prefixed) with the moonstone- prefix.

PR #164, opened 2020-08-20
MOON-140: Converts Typography to Typescript

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-140

## Description

* Converts Typography component to typescript and removes unnecessary children prop type check
* Changes usage of typescript in Badge component to be more consistent


src/components/Badge/Badge.tsx
```diff
@@ -3,57 +3,67 @@ import styles from './Badge.scss';
...
      * Badge label, only for type round
      */
-    label?: string,
+    label?: string;
 
     /**
      * Badge color
      */
-    color?: BadgeColor,
+    color?: TBadgeColor;
```
tdraier, line 30: Why can't we use directly `color?: BadgeColor` and `type?: BadgeType` ?

vindhya: I noticed when using enums for the type, the editor suggestions for the type don't show the values available (e.g., accent or success or danger), it would just show the type "BadgeColor" which isn't super useful.

Once I changed it to a type, the values did appear in the editor popup.

tdraier (approved):

src/components/Badge/Badge.tsx
vindhya: Screen Shot 2020-08-20 at 10 50 18 AM
Screen Shot 2020-08-20 at 11 04 29 AM

PR #177, opened 2020-09-16
Improved template to match the definition of ready of a component

Yousria:


ffffffelix (approved):

ffffffelix (approved):

.github/pull_request_template.md
```diff
@@ -33,14 +33,17 @@ If you need to provide more details, please use the description section.
 -->
 
 - [ ] All files present ( component - md - scss - spec - stories )
-- [ ] Are all props ok and documented
-- [ ] Required props and default values
+- [ ] All props ok and documented (propTypes)
```
vindhya, line 36: I think we should say `(propTypes or typescript types)` as we're in the process of moving components to typescript.

.github/pull_request_template.md
```diff
@@ -33,14 +33,17 @@ If you need to provide more details, please use the description section.
...
+- [ ] Allows custom props
 
 ## Documentation
 
 <!--
 Indicate if you have been writing documentation has part of this change.
 -->
 
-- [ ] README documentation
+- [ ] README documentation (Where/when and how the component should be used, structure, responsiveness, dependencies on other components, states, behaviours, animations, accessibility)
```
vindhya, line 48: The where/when and how the component should be used is a part of the figma component documentation right now as it's coming from Design.

Also, what is meant by "structure" here?

.github/pull_request_template.md
vindhya, line 36: I think we should also mention typescript types here as we're in the process of moving to typescript which means that propTypes aren't being used for the components which are in typescript.
```suggestion
- [ ] All props ok and documented (propTypes/typescript types)
```

.github/pull_request_template.md
rknj: OK with you where/when statement.

Looking at https://docs.google.com/document/d/1moEAolz106XH3XWeYH4-C6Zse2MfAVa5_ywZON_CjY8/edit, I think the idea behind structure is basic specification of the component

vindhya: Ok 👍🏾

vindhya (approved):

PR #178, opened 2020-09-16
MOON-135: Design updates to Input component

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-135

## Description

* Design updates for the Input component
* Cancel icon only appears if `onClear` prop is passed in AND the input is filled


ffffffelix (changes requested):
Please, can you just add `color: var(--color-gray);` to `input:disabled` to have a better contrast when the input is disabled. Thanks!

ffffffelix (approved):

PR #182, opened 2020-09-28
MOON-156: Migrates MenuItem to Typescript and Adds Images

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-156

## Description

* Migrates MenuItem component to Typescript
* Adds the ability to provide `small` or `big` images to MenuItem
    * The way the width and height are controlled need to be updated on the Menu component in the case that images are provided
* Adds the ability to provide `small` or `big` images to ListItem
* Additional fixes for ListItem:
    * Updates `index.js` to `index.ts` so it can be imported into Typescript files
    * Adds Moonstone truncate styling for when text is too long
    * Required explicit addition of `tabIndex` prop for Typescript compiler

## Checklist

- [x] All files present (component - md - scss - spec - stories)
- [x] All props ok and documented (propTypes or typescript types)
- [x] Required props, default values, variants and colors
- [x] Example in storybook

## Documentation

- [x] Design mockups (figma - where/when/how the component should be used): https://www.figma.com/file/QRmXslFDerZ7PBtRJecRbf/DESIGN-778?node-id=506%3A5149


src/components/Menu/MenuItem.scss
```diff
@@ -45,4 +50,26 @@
...
+
+    &.moonstone-menuItem-withImage_small {
+        height: $menuItem-withImage-height_default;
+        padding-left: 16px;
+    }
+
+    &.moonstone-menuItem-withImage_big {
+        height: $menuItem-withImage-height_big;
+        padding-left: 16px;
+    }
```
ffffffelix, line 62: You can use spacing tokens for padding
```suggestion
    &.moonstone-menuItem-withImage_small {
        height: $menuItem-withImage-height_default;
        padding-left: var(--spacing-medium);
    }

    &.moonstone-menuItem-withImage_big {
        height: $menuItem-withImage-height_big;
        padding-left: var(--spacing-medium);
    }
```

src/components/ListItem/ListItem.scss
```diff
@@ -1,4 +1,9 @@
+@import '../../utils/index';
+
 .moonstone-listItem {
+    @include truncate;
+
+    width: 100%;
     padding: 4px 8px;
```
ffffffelix, line 7: I know it's out of scope, but one day we will have to use token spacing here:
```suggestion
    padding: var(--spacing-nano) var(--spacing-small);
```

src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * MenuItem variants
+     */
+    variant?: TMenuItemVariant,
+
+    /**
+     * Optional image to display to describe the menu item
+     */
+    image?: HTMLImageElement,
```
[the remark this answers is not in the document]
vindhya: That could be okay - the reason I was thinking to use an `image` element is so that the user could provide alt text for the image. However, we could just use the label provided for the alt text as well.

src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+                'moonstone-selected': isSelected,
+                'moonstone-disabled': isDisabled,
+                'moonstone-title': variant === MenuItemVariant.TITLE
+            },
+            'flexRow_nowrap',
+            'alignCenter',
+            imageSize && `moonstone-menuItem-withImage_${imageSize}`
+        )}
+    >
+        <div className={`moonstone-menuItem-image_${imageSize}`}>{image}</div>
```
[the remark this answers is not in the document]
vindhya: Okay! Thanks!

src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+        <ListItem
```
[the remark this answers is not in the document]
vindhya: Good point, it's not the most correct HTML. How do you think we should wrap this component otherwise? Something needs to contain the image and the list item.

src/components/Menu/MenuItem.scss
```diff
@@ -45,4 +50,26 @@
...
+    .moonstone-menuItem-image_small img {
+        width: $default-image-size;
+        height: $default-image-size;
+    }
+
+    .moonstone-menuItem-image_big,
+    .moonstone-menuItem-image_big img {
+        width: $big-image-size;
+        height: $big-image-size;
+    }
```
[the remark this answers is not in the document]
vindhya: Ah, thank you, I forgot about the case of non-square images. Fixed now.

src/components/Menu/MenuItem.tsx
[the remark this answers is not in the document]
ffffffelix: Good point, we can keep it like this

src/components/ListItem/ListItem.scss
```diff
@@ -1,7 +1,38 @@
...
+    @include truncate;
+
+    width: 100%;
 
     list-style: none;
+
+    .moonstone-listItem-image_small {
+        flex: 0 0 $small-image-size;
+        height: $small-image-size;
+        margin-right: var(--spacing-small);
```
ffffffelix, line 16: It's not come from your PR, but it miss a spacing between the icon and the label. We always must have a space between the image/icon and the text. So you can add this margin to the Typography element, maybe you can add a class for that.

src/components/ListItem/ListItem.tsx
```diff
@@ -46,14 +69,21 @@ export const ListItem: React.FC<IListItemProps> = ({
...
             {...props}
         >
             {
-            iconStart &&
+                iconStart &&
                 <iconStart.type {...iconStart.props} size="small" className={classnames('moonstone-listItem_iconStart')}/>
             }
 
+            {
+                image &&
```
ffffffelix, line 83: Do we prevent the possibility to have an icon and an image at the same time with something like `image && !iconStart` ?

src/components/ListItem/ListItem.tsx
```diff
@@ -46,14 +69,21 @@ export const ListItem: React.FC<IListItemProps> = ({
...
-        <li className={classnames(cssListItem)}
+        <li
+            className={classnames(cssListItem)}
+            tabIndex={tabIndex}
             {...props}
         >
             {
-            iconStart &&
+                iconStart &&
                 <iconStart.type {...iconStart.props} size="small" className={classnames('moonstone-listItem_iconStart')}/>
```
ffffffelix, line 79: It seems the className `moonstone-listItem_iconStart` render `undefined` . You can see it in the `ListItem/Icon + Text` storybook example.

src/components/ListItem/ListItem.tsx
```diff
@@ -65,7 +95,7 @@ export const ListItem: React.FC<IListItemProps> = ({
             </Typography>
 
             {
-            iconEnd &&
+                iconEnd &&
                 <iconEnd.type {...iconEnd.props} size="small" className={classnames('moonstone-listItem_iconEnd')}/>
```
ffffffelix, line 99: Same here the classname `moonstone-listItem_iconEnd` isn't display

src/components/ListItem/ListItem.stories.jsx
```diff
@@ -13,25 +13,43 @@ storiesOf('Components|ListItem', module)
         notes: {markdown: markdownNotes}
     })
     .addDecorator(withKnobs)
-    .add('default', () => (
+    .add('Default', () => (
         <div style={{transform: 'scale(1)', height: '100vh'}}>
```
ffffffelix, line 17: Just to be more semantic, can we change these `div`s to an `ul` ?

src/components/ListItem/ListItem.tsx
```diff
@@ -46,14 +69,21 @@ export const ListItem: React.FC<IListItemProps> = ({
...
     return (
-        <li className={classnames(cssListItem)}
+        <li
+            className={classnames(cssListItem)}
+            tabIndex={tabIndex}
             {...props}
         >
             {
-            iconStart &&
+                iconStart &&
```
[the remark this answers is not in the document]
vindhya: Yes, we can do that. It could be confusing for the user to have both the `iconStart` and the `image` disappear if both are specified but I can make a note in the documentation of the props.

src/components/ListItem/ListItem.tsx
vindhya: Same answer as above for `iconStart && !image` - could be confusing for the user but I'll add info to the props documentation.

src/components/ListItem/ListItem.scss
vindhya: There was a class that was supposed to be applying margin to the left or right of the icon but it wasn't appearing - changed structure a bit and put the icon inside a div so that the margin is applied - need to investigate why the custom classes are showing up as `undefined` for the Icon components later.

src/components/ListItem/ListItem.tsx
vindhya: Ah yes, I saw this as per above. I've made a fix for now.

src/components/ListItem/ListItem.tsx
vindhya: The className should appear now.

src/components/ListItem/ListItem.stories.jsx
vindhya: Done!

ffffffelix (approved):

PR #191, opened 2020-10-22
MOON-201: Fixes issue with searching when Dropdown Menu has groups

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-201

## Description

* Errors occurred when searching within Dropdown Menus that had grouped options
* Added a function within Menu component to aggregate searchable children


ffffffelix:
The `searchEmptyText` isn't displayed anymore if there is no search result.

ffffffelix (approved):

PR #192, opened 2020-10-27
MOON-164: Adds tests for new Dropdown and Menu features

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-164

## Description

* Added react testing library and associated packages
* Converts tests in Dropdown and Menu to react testing library and added tests for new features
* Added `isHtml` prop with default value `false` back into list for Typography to fix DOM errors
* Small fix for search empty text showing up when it shouldn't


src/components/Dropdown/Dropdown.tsx
```diff
@@ -252,6 +252,7 @@ export const Dropdown: React.FC<IDropdownProps> = (
         <div
             className={classnames(className)}
             style={{maxWidth}}
+            data-testid="moonstone-dropdown"
```
ffffffelix, line 255: As the purpose of this attribute is to add a selector for test, I propose adding it only on the test.

src/components/Dropdown/Dropdown.spec.js
```diff
@@ -18,32 +19,73 @@ const data = [
     }
 ];
 
+const dataGrouped = [
```
ffffffelix, line 22: What do you think to put all data in `src/data` to avoid repeating the same fake data object for storybook and tests (like we did for the TreeView)?

src/components/Menu/Menu.tsx
```diff
@@ -228,6 +228,7 @@ export const Menu: React.FC<IMenuProps> = (
                 )}
                 onMouseEnter={onMouseEnter}
                 onMouseLeave={onMouseLeave}
+                data-testid="moonstone-menu"
```
ffffffelix, line 231: Same here, can we add this attribute for test only?

ffffffelix:
We haven't 100% of the component tested. Is it something we want to achieve?
Dropdown features not tested for now:
- The size `small` or `medium`
- The icon 
- `handleOpenMenu`
- `handleSelect`
- `handleCloseMenu`
- `handleKeyPress`

src/components/Dropdown/Dropdown.tsx
vindhya: Good point! It's not necessary for it to be in the component itself. I'll move it to the test.

src/components/Dropdown/Dropdown.spec.js
vindhya: I think we need some of the test data in the story itself, otherwise, it will be difficult for developers implementing the component to see what the data format should be - they would have to leave storybook and go to github to check the source code.

src/components/Menu/Menu.tsx
vindhya: Yes!

vindhya:
> We haven't 100% of the component tested. Is it something we want to achieve?
> Dropdown features not tested for now:
> 
> * The size `small` or `medium`
> * The icon
> * `handleOpenMenu`
> * `handleSelect`
> * `handleCloseMenu`
> * `handleKeyPress`

As these weren't already tested, I left them out but it could be good to test them in the future. I'll create a Moonstone ticket so that we can remember to do it.

ffffffelix (approved):

PR #202, opened 2020-11-18
MOON-153: Set up CI/CD

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-153

## Description

* Set up CircleCI so that for linting and testing on branches
* CircleCI runs the release process when changes are made to the master branch (e.g., a PR is merged to `master` from `develop` when a release is necessary)


.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
+version: 2.1
+orbs:
+  node: circleci/node@4.1.0
```
Fgerthoffert, line 3: the circleci images are being deprecated, you might want to use cimg instead

https://circleci.com/developer/images/image/cimg/node

PS: we need to update config files in the other repos as well

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+        - .
+
+  attach-workspace: &attach-workspace
+    attach_workspace:
+      at: .
+
+defaults: &defaults
+  working_directory: ~/source
+  docker:
+    - image: circleci/node:latest
```
Fgerthoffert, line 19: same comment regarding cimg images

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+    <<: *defaults
+    steps:
+      - *attach-workspace
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
+
+      - run: yarn install
```
Fgerthoffert, line 58: why yarn install again? It's already in the lint-test job

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+      - *attach-workspace
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
+      
+      - run:
+          name: Setup .env
+          command: echo -e "GH_TOKEN=${CIRCLECI_MOONSTONE_RELEASE_TOKEN}\nNPM_TOKEN=${NPMJS_TOKEN}" > ~/source/.env
```
Fgerthoffert, line 78: It's super minor but might improve a bit readibility if you were to write it that way:
command: |
  echo -e "GH_TOKEN=${CIRCLECI_MOONSTONE_RELEASE_TOKEN} > ~/source/.env
  echo -e "NPM_TOKEN=${NPMJS_TOKEN} >> ~/source/.env
 ```

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+      - run:
+          name: Authenticate with registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPMJS_TOKEN" > ~/source/.npmrc
+      - run: yarn install
+      - run:
+          name: Run build
+          command: yarn build
```
Fgerthoffert, line 85: Why running yarn build again ? It's already in the build job

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+      - run:
+          name: Release and publish
+          command: |
+            git config user.email "jahia-ci@jahia.com"
+            git config user.name "Jahia CI"
+            git fetch --tags
+            yarn auto-changelog
+            yarn auto-version
+            npm publish --access public
+            git push --follow-tags --set-upstream origin master
```
Fgerthoffert, line 96: Out of curiosity, this does not trigger another circleci run since you're pushing something back to master ?

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+            - v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
+
+      - run:
+          name: Install dependencies
+          command: yarn install
+      - run:
+          name: Run linting
+          command: yarn lint
```
Fgerthoffert, line 38: --max-warning 1 ?

Fgerthoffert (approved):

.circleci/config.yml
vindhya: I tried to use the new cimg docker image but was getting errors with node-sass so I reverted back to the old circleci node docker image 😞

Fgerthoffert: good to know, with the same version of the image ?

vindhya: With `cimg/node:14.15` since it doesn't seem like the `latest` tag exists for the new docker images.

.circleci/config.yml
vindhya: I was thinking just in case 😛 But, you're correct, it's redundant so I'll remove it.

.circleci/config.yml
vindhya: Nice! I like that better 😄  Will make the change.

.circleci/config.yml
vindhya: I wasn't sure whether the build files would carry over but I guess they would since the cache is being persisted 👍🏾

.circleci/config.yml
vindhya: The commit that's generated by auto release has the [skip ci] in the commit messages.

PR #207, opened 2020-12-04
MOON-216: Fixes menu truncation issue in Firefox and Safari

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-216

## Description

* The `Menu` opened by `Dropdown` will have a width which is no smaller than 70px
* Removed the node options to increase memory for node as they're no longer necessary with the storybook 6 update

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "react": "^16.11.0"
     },
     "scripts": {
-        "start": "NODE_OPTIONS=--max_old_space_size=4096 start-storybook -p 6006",
+        "start": "start-storybook -p 6006",
         "test": "jest --ci --runInBand --coverage",
         "tdd": "jest --watch",
         "lint:js": "eslint --ext js,jsx,json .",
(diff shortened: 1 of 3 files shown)


ffffffelix:
Instead of adding new props to manage this case, I propose to add a new minimal constant to don’t exceed (anyway below this value the menu is not very usable)
`const menuMinWidth = 70;`

And when we set the min-width, checking if the calculated min-width doesn’t exceed our minimal width. To be sure to have a usable menu when the dropdown is very small.
```
const dropdownWidth = (e.currentTarget as HTMLElement).offsetWidth;
setMinWith(dropdownWidth > menuMinWidth ? dropdownWidth : menuMinWidth);
```

What do you think?

vindhya:
> Instead of adding new props to manage this case, I propose to add a new minimal constant to don’t exceed (anyway below this value the menu is not very usable)
> `const menuMinWidth = 70;`
> 
> And when we set the min-width, checking if the calculated min-width doesn’t exceed our minimal width. To be sure to have a usable menu when the dropdown is very small.
> 
> ```
> const dropdownWidth = (e.currentTarget as HTMLElement).offsetWidth;
> setMinWith(dropdownWidth > menuMinWidth ? dropdownWidth : menuMinWidth);
> ```
> 
> What do you think?

I would be okay with that, but if the user had a valid reason to want a super small Menu, maybe we should provide a way to turn off this functionality. Instead of having the prop to turn it on, what do you think about having it work with the 70px minimum width by default and then having a prop to turn it off and use the minimum width provided by the user?

ffffffelix:
Mwep, we can do that even if I don't see any valid reason to have super tiny menu's width in the dropdown.

vindhya:
Okay..if you don't see any valid reason, then I'll remove the flag and have 70px as the ultimate minimum width for the dropdown's menu. I don't want to complicate the component unnecessarily 🙂

PR #208, opened 2020-12-07
MOON-217:  Update dependencies

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-217

## Description
- Update Dependencies
- Update DevDependencies
- Remove `storybook-addon-versions`, `storybook-readme` and `"@storybook/addon-links` because we don't use them.
- Add `eslint-plugin-react-hooks` to be compatible with the last eslint. (Shouldn't be load from `@jahia/eslint-config` ? )
- Remove unnecessary rules in the storybook's  webpack config 
- Adapt stories to be compatible with storybook 6


vindhya (approved):

vindhya:
Yeah...@jahia/esling-config has eslint-plugin-react-hooks so I'm not sure why we're installing it again here.

ffffffelix:
> Yeah...@jahia/esling-config has eslint-plugin-react-hooks so I'm not sure why we're installing it again here.

Following the eslint documentation : https://eslint.org/docs/developer-guide/shareable-configs, the plugins should be in `peerDependency ` and loaded in each project. But I don't understand why we didn't have the warning before.

PR #218, opened 2020-12-30
MOON-220: Avoid to render something when the component is empty

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-220

## Description

Avoid rendering something when the component is empty and have no children/items.

## Tests

- [X] Unit Test

### DIFF
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -51,6 +51,11 @@ export const Badge: FunctionComponent<BadgeProps> = ({
         `moonstone-${type}`,
         className
     );
+
+    if (!label || label.length < 1) {
+        return null
+    }
+
     if (type === BadgeType.Round) {
         return (
             <Typography isNowrap component="span" variant="caption" weight="bold" className={classNameProps} {...other}>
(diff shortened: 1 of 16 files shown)


ffffffelix:
Some warnings are displayed during tests because I have to don't pass some required props to verify if the component displays nothing when it has no children.
Does someone have an idea to prevent that?

src/components/Badge/Badge.tsx
```diff
@@ -51,13 +51,19 @@ export const Badge: FunctionComponent<BadgeProps> = ({
         `moonstone-${type}`,
         className
     );
+
+    if (!label || label.length < 1) {
```
vindhya, line 55: Can't the diamond type of the Badge be used without a label?

ffffffelix: You right!
I'm wondering if this variant is used, I'd like to remove it.

vindhya: I haven't seen it being used: https://github.com/search?q=org%3AJahia+type%3D%22diamond%22&type=code

vindhya (approved):

PR #220, opened 2021-01-21
TECH-336: Migrate Button component to TS and other refactors

vindhya:
## JIRA

https://jira.jahia.org/browse/TECH-336


src/__storybook__/utils.ts
```diff
@@ -0,0 +1,14 @@
+import * as Icons from '~/icons';
+export const iconsName = Object.keys(Icons);
+
+export function capitalize(str: string) {
```
ffffffelix, line 4: Does that mean we can export SVG icons without the first letter capitalized?
It's cool in case we forget to capitalize the icon during the export from Figma.

ffffffelix (commented):
I don’t understand the value to have the type definition in a `shared-types` folder instead of the component folder?

vindhya:
> I don’t understand the value to have the type definition in a `shared-types` folder instead of the component folder?

Initially, I thought shared types should be a level higher to avoid dependency between components but, after thinking about it more, since dependencies already exist and because our dist has plain javascript as well now (not just bundled/minified/etc), I think it's okay to move them back to the component folders. I'll update the PR.

src/__storybook__/utils.ts
vindhya: I don't think so...it's just an array of the icon names that are available - I didn't add anything here. For some reason, git thinks it's a new file instead of a rename of the utils.js to utils.ts.

ffffffelix (approved):

PR #226, opened 2021-02-08
TECH-342: Migrate ResizableBox to Typescript

vindhya:
## JIRA

https://jira.jahia.org/browse/TECH-342

## Description

* Migrated ReizableBox to typescript
* Added typescript files to .eslintignore because, currently, we are using TSLint to lint typescript files until the typescript migration is complete. At which point, we will only use ESLint


src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,58 @@
...
+}) => {
+    const enableZones: EnableZones = {};
+
+    zones.forEach(zone => {
+        enableZones[zone] = enable.indexOf(zone as any) > -1;
+    });
+
+    return (
+        <Resizable
+            data-role="resizable-panel"
```
ffffffelix, line 29: Adding the `role` type  definition on `ResizableBoxProps` seems fix the typescript error. In that way, we avoid some breaking changes for our selenium tests.

vindhya: Ahh, nice. I'm not sure why that works but it does! I've made the update 😄

ffffffelix (approved):

PR #227, opened 2021-02-10
MOON-264: Create the Loader components

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-264

## Description

- Create a loader component with 3 different sizes
- Add  `@storybook/addon-controls`

## Tests

- [X] Unit Test skeleton

## Checklist

- [X] All files present (component - md - scss - spec - stories)
- [X] All props ok and documented (propTypes or typescript types)
- [X] Required props, default values, variants and colors
- [X] Example in storybook
- [X] Allows custom props

## Documentation

- [ ] README documentation (specification of the component, responsiveness, dependencies on other components, states, behaviours, animations, accessibility)
- [ ] Design mockups (figma - where/when/how the component should be used)

### DIFF
--- a/src/components/Loader/Loader.stories.tsx
+++ b/src/components/Loader/Loader.stories.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import {Story} from '@storybook/react'
+import {Loader} from './index';
+import {LoaderProps} from './Loader.types';
+
+export default {
+    title: 'Components/Loading',
+    component: Loader
+};
+
+const Template: Story<LoaderProps> = args => <Loader {...args}/>;
+
+export const Loading = Template.bind({});
+Loading.args = {
+    size: 'small'
+};
--- a/src/components/Loader/Loader.stories.tsx
+++ b/src/components/Loader/Loader.stories.tsx
@@ -1,16 +1,16 @@
 import React from 'react';
 import {Story} from '@storybook/react'
-import {Loader} from './index';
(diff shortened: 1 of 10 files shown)


src/components/Loader/Loader.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+    title: 'Components/Loading',
+    component: Loader
+};
+
+const Template: Story<LoaderProps> = args => <Loader {...args}/>;
+
+export const Loading = Template.bind({});
+Loading.args = {
```
vindhya, line 14: If you install `@storybook/addon-controls` and add it to the list of addons in main.js, then the controls should show up in the story.

ffffffelix: Can we have `addon-knobs` and `addon-controls` to the same project ? Or do we prefer to migrate all stories to `addon-essentials` at the same time ?

vindhya: I think it could be good to slowly start migrating to `addon-controls` when we can and turn `addon-knobs` off for those components which have controls.

Later, when everything has been changed to controls, we can remove `@storybook/addon-knobs`, `@storybook/addon-controls`, and add `@storybook/addon-essentials`. Sound okay?

ffffffelix: I added `@storybook/addon-controls`

vindhya (approved):

PR #230, opened 2021-02-19
MOON-270: Rename classnames to clsx

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-270

## Description
- Use `clsx` everywhere we were still using `classnames` to add classes to some stories
- Replace `classnames` to `clsx`

### DIFF
--- a/src/tokens/colors/color.stories.jsx
+++ b/src/tokens/colors/color.stories.jsx
@@ -1,11 +1,9 @@
 import React from 'react';
 import {storiesOf} from '@storybook/react';
 import PropTypes from 'prop-types';
-import classnames from 'classnames/bind';
+import clsx from 'clsx';
 import styles from './colors.stories.module.scss';
 import storyStyles from '~/__storybook__/storybook.module.scss';
-
-let cx = classnames.bind(styles);
 
 const paletteNeutral = ['light',
     'light60',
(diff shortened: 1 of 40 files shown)


src/tokens/colors/color.stories.jsx
```diff
@@ -52,9 +50,9 @@ const paletteColors = [
 
 export const Color = ({color, name}) => {
     return (
-        <div className={classnames(storyStyles.storyItem)}>
+        <div className={clsx(storyStyles.storyItem)}>
             <p>{name}</p>
-            <div className={classnames(styles.colorShape, cx([`color-${color}`]))}/>
+            <div className={clsx(styles.colorShape, styles[`color-${color}`])}/>
```
vindhya, line 55: What is `cx`/`styles` here?

vindhya (approved):

src/tokens/colors/color.stories.jsx
ffffffelix: `cx` referred to `let cx = classnames.bind(styles);` 
`styles ` referes to `import styles from './colors.stories.module.scss';

It's to load a CSS from a story, it's seems to just import the scss file doesn't work here.

vindhya: Ahh gotcha 👍🏾

PR #234, opened 2021-03-05
TECH-339: Migrate GlobalStyle, ImgWrapper, and SvgWrapper components to Typescript

vindhya:
## JIRA

https://jira.jahia.org/browse/TECH-339

## Description

Components migrated to Typescript:
* GlobalStyle
* ImgWrapper
* SvgWrapper
   * Changed the way the `viewBox` and `fill` attributes are accessed because the `.attributes` way is an old API and causes issues with types


src/components/ImgWrapper/ImgWrapper.types.ts
```diff
@@ -0,0 +1,18 @@
+export type ImgWrapperSize = 'small' | 'default' | 'big';
+export const imgWrapperSize = ['small', 'default', 'big'];
```
ffffffelix, line 2: It seems for other components (Button, Badges, Chip ...) we use the pluriel to describe the array `imgWrapperSizes`.
I don't know which convention we want to apply.

src/components/SvgWrapper/SvgWrapper.types.ts
```diff
@@ -1,6 +1,9 @@
-import * as React from 'react';
+export type SvgWrapperSize = 'small' | 'default' | 'big';
+export const svgWrapperSize = ['small', 'default', 'big'];
```
ffffffelix, line 2: Same question here, do we want to use the pluriel: `svgWrapperSizes`?

src/components/ImgWrapper/ImgWrapper.types.ts
vindhya: I updated it to be the same as the other components. Nice catch 👀

src/components/SvgWrapper/SvgWrapper.types.ts
vindhya: Updated!

ffffffelix (approved):

PR #235, opened 2021-03-05
TECH-343: Migrate SecondaryNav and SecondaryNavHeader to Typescript

vindhya:
## JIRA

https://jira.jahia.org/browse/TECH-343

## Description
* Migrated SecondaryNav and SecondaryNavHeader components to typescript


tslint.json
```diff
@@ -7,7 +7,8 @@
     "rules": {
         "jsx-boolean-value": [true, "never"],
         "jsx-wrap-multiline": true,
-        "jsx-no-lambda": false
+        "jsx-no-lambda": false,
+        "no-empty": false
```
ffffffelix, line 11: Why do you want to set the `no-empty` rule to `false` ?

ffffffelix:
If we want to keep the rule, what do you think to use `() => undefined` instead of `() => {}` ?

ffffffelix (approved):

PR #238, opened 2021-03-10
TECH-333 : move Accordion components to typeScript

dgriffon:
https://jira.jahia.org/browse/TECH-333


src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -61,53 +64,22 @@ export const AccordionItem = ({id, label, icon, onClick, children, className, ..
...
                     {children}
-                </div>}
+                </div>
+                )}
         </section>
     );
 };
 
 AccordionItem.defaultProps = {
     icon: null,
```
vindhya, line 82: As these default props are being used explicitly in the component, you can try moving to the component function definition as default parameters, e.g., `...: React.FC<AccordionItemProps> = ({id = null, ...` on line 9.

However, I've noticed that this, sometimes, causes issues with the component rendering. If so, please keep it the way it is with the `defaultProps` property.

src/components/Accordion/ControlledAccordion.tsx
```diff
@@ -0,0 +1,38 @@
...
+                }
+                 {...props}
+            >
+                {children}
+            </div>
+        </AccordionContext.Provider>
+    );
+};
+
+ControlledAccordion.defaultProps = {
```
vindhya, line 32: Same comment about these default props as the comment in AccordionItem.tsx.

src/components/Accordion/ControlledAccordion types.ts
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
vindhya, line 1: I think this file name is missing a `.` 🙂

src/components/Accordion/UncontrolledAccordion types.ts
```diff
@@ -0,0 +1,24 @@
+import React, {useState} from 'react';
```
vindhya, line 1: I think this file name is also missing a `.` 🙂

vindhya (approved):

PR #239, opened 2021-03-11
TECH-349: Typescript format cleanup

vindhya:
## JIRA

https://jira.jahia.org/browse/TECH-349

## Description

* Updated typescript format to be consistent for all components
* Upgraded Storybook version


src/components/Typography/Typography.tsx
```diff
@@ -74,7 +26,7 @@ export const Typography: FunctionComponent<TypographyProps> = ({
     isNowrap = false,
     isHtml = false,
     ...props
-}: TypographyProps) => {
```
ffffffelix, line 77: Is it ok to remove it ? I don't really understand what it does, I just notice sometimes we have it (Button, ButtonGroup, Chip, Loader, Tab) sometimes not.

ffffffelix:
Nice work @vindhya ! 🙌

src/components/Typography/Typography.tsx
vindhya: Yes, it's okay to remove it after the props. I was noticing that the linter was complaining about it but it might have been ESLint conflicting with TSLint. Now that ESLint is ignoring ts/tsx files, I'm going to check whether it can be removed in all of those components.

vindhya: I've removed the unnecessary types from those components now. Would you mind taking one more quick look at the changes?

ffffffelix (approved):

PR #240, opened 2021-03-19
MOON-256: Create the Header component

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-256

## Description

- Create the Header component

## Tests

- [X] Unit Test skeleton

## Checklist

- [X] All files present (component - md - scss - spec - stories)
- [X] All props ok and documented (typescript types)
- [X] Required props, default values, variants and colors
- [X] Example in storybook
- [X] Allows custom props

## Documentation

- [ ] README documentation (specification of the component, responsiveness, dependencies on other components, states, behaviours, animations, accessibility)
- [X] Design mockups (figma - where/when/how the component should be used)

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -17,14 +17,13 @@ export const Header: React.FC<HeaderProps> = ({
     backButton = null,
     breadcrumb,
     className,
-    ...other
+    ...props
 }: HeaderProps) => {
     const hasInformationArea = Boolean(breadcrumb || contentType || status);
     const hasToolbar = Boolean(toolbarLeft || toolbarRight);
 
     return (
-        <header className={clsx('moonstone-header', className)} {...other}>
-
+        <header className={clsx('moonstone-header', className)} {...props}>
             {/* Main area */}
             <div className={clsx('moonstone-header_main', 'flexRow', 'alignCenter', 'flexFluid')}>
                 { backButton && (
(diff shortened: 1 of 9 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    <div className={clsx('moonstone-header_mainActions', 'moonstone-header_actions')}>
+                        { mainActions }
+                    </div>
+                )}
+            </div>
+
+            { hasInformationArea && (
+                <div className={clsx('flexRow_between', 'alignCenter', 'moonstone-header_information')}>
+                    <div className={clsx('flexRow_nowrap', 'alignCenter', 'flexFluid', 'moonstone-header_informationLeft')}>
+                        { breadcrumb }
```
tdraier, line 56: Isn't this very specific to a type of header editing content ? Don't we have use cases where a more generic "information area" could be required ?

vindhya: I guess it depends if we want to restrict to this type of information area...

vindhya (approved):

src/components/Header/Header.tsx
ffffffelix: From what I saw, we don't have other example for an information area needs. The idea is to facilitate the implementation to manage spacing, separator from the component itself. Plus if we want to modify where the elements are positioned it will not impact the implementation.

tdraier (approved):

PR #244, opened 2021-05-04
MOON-190: WIP - Table Components

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-190

## Description

The Table components are a work in progress. This PR is to avoid having to submit an even larger PR in the future.


src/components/Table/TableCell/TableCell.tsx
```diff
@@ -0,0 +1,39 @@
...
+        {
+            className: clsx(
+                'moonstone-table_cell',
+                'horizontalAlign' + capitalize(horizontalAlign),
+                'verticalAlign' + capitalize(verticalAlign),
+                className
+            ),
+            ...props
+        },
+        component === 'th'
```
ffffffelix, line 30: I'm wondering if we should split it into two components `TableCellHead` and `TableCellBody`. the `th`  element has many differences (typography, sorting, no hover/selection ....) whereas the `td` element is just a wrapper. What do you think ?

src/utils/helpers.ts
```diff
@@ -0,0 +1,3 @@
+// is this folder only supposed to be for scss helpers?
```
ffffffelix, line 1: Nope, it's fine to have javascript too. Maybe we can organize this folder `utils/styles` and `utils/scripts` ? It depends if we want to put all utilities scripts in the same file or not.

src/components/Table/TableRow/TableRow.scss
```diff
@@ -0,0 +1,21 @@
+.moonstone-table_row {
+    border-bottom: 1px solid var(--color-gray40);
+
+    &:hover {
```
ffffffelix, line 4: We have to prevent the hover for the row in the table header.

src/components/Table/TableRow/TableRow.types.ts
```diff
@@ -0,0 +1,23 @@
...
+export type TableRowProps = {
+    /**
+     * Any additional class names to apply
+     */
+    className?: React.ReactNode;
+
+    /**
+     * What type of spacing the cell should have
+     */
+    height?: 'regular' | 'relaxed';
```
ffffffelix, line 12: I'm not convinced with `regular` and `relaxed` but medium/big ... is not better. So I don't have a better idea.

src/components/Table/TableCell/TableCell.scss
```diff
@@ -0,0 +1,30 @@
+.moonstone-table_cell {
```
ffffffelix, line 1: I think it's better to use the same name than the component but in camelCase: `moonstone-tableCell`.

src/components/Table/TableRow/TableRow.scss
```diff
@@ -0,0 +1,21 @@
...
+        background-color: var(--color-gray_light40);
+    }
+
+    &.moonstone-table_row-selected {
```
ffffffelix, line 8: This comment is for all CSS class names. I propose to use the same name as the component but in camelCase and use underscore for modifiers like: `moonstone-tableRow_selected`. 
I know we don't have a real naming convention for CSS (yet). What do you think?

src/components/Table/Table.scss
```diff
@@ -0,0 +1,4 @@
+// I didn't need to add this class to the component?!
+.moonstone-table {
```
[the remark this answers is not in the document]
vindhya: I'll remove this scss file, but the crazy thing is that this class was being applied to the root of the Table component in the DOM even though I never specifically applied it! 🤯

src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import {
+    Table,
+    TableHead,
+    TableRow,
+    TableBody,
+    TableCell
+} from '~/components';
+import {tableDataFlat} from '~/data/tableDataFlat';
+import './Table.scss';
```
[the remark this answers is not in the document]
vindhya: Ah yes! That was a mistake. Thanks 🙂

src/components/Table/TableBody/TableBody.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import {TableBodyProps} from './TableBody.types';
+
+export const TableBody: React.FC<TableBodyProps> = ({children}) => (
+    <tbody>
```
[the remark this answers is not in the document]
vindhya: Yes! This was just me not double-checking.

src/components/Table/TableCell/TableCell.scss
```diff
@@ -0,0 +1,30 @@
...
+    padding: 0 var(--spacing-medium);
+}
+
+// is there a better place to put these utility classes? maybe somewhere in the global scss
+// will text align be okay for all types of content in the cell?
+// refactor into scss @mixin or @function or something
+.horizontalAlignLeft {
```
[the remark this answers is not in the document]
vindhya: I'm going to keep them as classes for now and move them to `src/globals`. I think `_alignment` is a pretty descriptive name and I can't think of anything better 🙂

src/components/Table/TableCell/TableCell.tsx
vindhya: Hmm yes, that could make the components simpler. I'm going to make `TableHeadCell` and `TableBodyCell`.

src/utils/helpers.ts
vindhya: As there's only 1 file of utility scripts, for now, I'll just keep this one helpers.ts file in the folder. Once we add more, we can break them into those folders.

src/components/Table/TableRow/TableRow.types.ts
vindhya: In Gmail, they have the naming convention, `default`, `comfortable`, and `compact`. We can think about it some more. It doesn't have to be finalized until we need to make the release.

src/components/Table/TableCell/TableCell.scss
vindhya: If we're using a CSSBEM-like naming convention, should it be like `moonstone-block-element_modifier`? Where table is the block in this case and cell is the element?

src/components/Table/TableRow/TableRow.scss
vindhya: Same comment as above...I think this should be `moonstone-table_row-selected`.

src/components/Table/TableRow/TableRow.scss
vindhya: I'm thinking about the most simple way to organize the `TableHead` and `TableRow` components with regard to this. I was thinking about having the `TableHead` component include a `tr` since it should always be used with a `tr` element anyway. But then, `TableHead` and `TableBody` would differ in how they're used, and, `TableHead` would differ a little from the way the HTML implementation looks as well - I'm not sure if this is actually an issue, though.

src/components/Table/TableRow/TableRow.types.ts
ffffffelix: In Gmail, it changes the padding top/bottom to change the density of each row whereas for us we just change the height because we want to display two lines in a cell but we keep the same density. 
Maybe `height?: oneLine | twoLine`. But it's very descriptive.
We can keep it like this for now, and discuss it during our next meeting.

src/components/Table/TableCell/TableCell.scss
ffffffelix: If we follow the BEM documentation you right, but I'd like to have the component name associate with the class name. I think it easier to debug/maintain.

ffffffelix (approved):

src/components/Table/TableRow/TableRow.types.ts
ffffffelix: After a discussion with Samson, as we want to avoid people adding the wrong height value for the single-line row. We think it can be clearer to replace the height property by: `isMultipleLine ?: boolean`. What do you think?

vindhya: That sounds good to me.

PR #245, opened 2021-05-17
MOON-285: Table row selection with react-table

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-285

## Description

The `Table.stories.jsx` file shows how row selection will work with `react-table` and Moonstone components.

An uncontrolled Checkbox component variant was created as a part of these changes to support `react-table`.


src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,37 +1,15 @@
...
-            </svg>
-        </div>
-    );
+import {UncontrolledCheckbox} from './UncontrolledCheckbox';
+import {ControlledCheckbox} from './ControlledCheckbox';
+
+export const Checkbox: React.FC<CheckboxProps> = ({variant = 'controlled', ...props}) => {
+    // Controlled is the default variant of checkbox to use because components should
+    // be controlled whenever possible
+    return variant === 'uncontrolled'
```
ffffffelix, line 10: It seems we use `variant` for differentiating style, I don't like to use it for technical purposes here. Do you think we can implement it in a transparent way as we did for the `TreeView` component?

src/components/Table/Table.stories.jsx
```diff
@@ -0,0 +1,234 @@
...
+                            {
+                                selectedRowIds,
+                                'selectedFlatRows[].original': selectedFlatRows.map(d => d.original)
+                            },
+                            null,
+                            2
+                        )}
+                    </code>
+                </pre>
+            </section>
```
ffffffelix, line 217: It will be good if we can add these kinds of information in storybook plugins like `storybook-actions` to display it in a dedicated tab instead of the main canvas.

src/components/Checkbox/UncontrolledCheckbox.tsx
```diff
@@ -0,0 +1,54 @@
...
+                        {...otherProps}
+                        {...focusProps}
+                    />
+                    <svg className={clsx('moonstone-checkbox_icon')} viewBox="0 0 21 21">
+                        { indeterminate
+                            ? <path d="M4.5 10.5L16.5 10.5" strokeLinecap="round"/>
+                            : <path d="M5 10.75L8.5 14.25L16 6" strokeLinecap="round"/>
+                        }
+                    </svg>
+                </div>
```
ffffffelix, line 48: Can we use the controlled version of the component as we did for the `TreeView` component to avoid duplicating code?

src/components/Checkbox/Checkbox.tsx
vindhya: I'm not sure what property we can check against to differentiate between to use the controlled vs uncontrolled version. Any of the properties could be present in either version, i.e., `checked`, `aria-checked`, etc. Do you think it could be a different prop name? Maybe something like `type`?

src/components/Checkbox/UncontrolledCheckbox.tsx
vindhya: Unfortunately,  the `react-table` library needs direct access to an uncontrolled version of the component to control. That's the only reason why I had to add this version.
Another approach we could take is to have the `Checkbox` component be controlled and the uncontrolled version only be accessed through a component called `UncontrolledCheckbox`. That way, only people who specifically want an uncontrolled version will use that component.

ffffffelix (approved):

src/components/Checkbox/Checkbox.tsx
vindhya: I'm changing this to be a boolean, `isUncontrolled`, where it is false by default.

src/components/Checkbox/UncontrolledCheckbox.tsx
vindhya: The structure `Checkbox` component will stay as it is for now. [MOON-293](https://jira.jahia.org/browse/MOON-293) has been created to refactor this in the future.

PR #246, opened 2021-05-20
MOON-283: Set up postcss-preset-env (includes autoprefixer)

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-283

## Description

- Set up webpack and build configs to use `postcss-preset-env`. This includes `autoprefixer`, which adds browser flags to css rules that may not be supported in all browsers, and lets us use newer stage 2+ css features by polyfilling them when necessary
- Had to use older versions of `sass-loader` and `postcss-loader` because there were conflicts


scripts/compile-css.js
```diff
@@ -0,0 +1,50 @@
...
+        let target = path.resolve('dist', outputFile);
+        let folder = path.resolve(target, '..');
+
+        if (!fs.existsSync(folder)) {
+            fx.mkdirSync(folder);
+        }
+
+        postcssPresetEnv.process(preCss.css)
+            .then(postCss => {
+                console.log(postCss.css);
```
ffffffelix, line 47: Can remove this log as it print all css :)

ffffffelix (approved):

scripts/compile-css.js
vindhya: Oops! Will do 😅

PR #248, opened 2021-06-02
MOON-294: Fix the icons ChevronBig and add new icons

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-294

## Description

- I updated the icons `ChevronBigLeft` and `ChevronBigRight` to have the same width
- I added the icons `ChevronFirstPage`, `ChevronLastPage` and `WorkInProgress`
- I renamed the SVG icon `Menu` by `MenuIcon` to prevent a name conflict with the `Menu` component during the build. Previously we updated the exported name in `src/icons/assets.ts` but this file is automatically generated by `yarn build-icons`. So when we run the command we lost the modification.

Note:
When I run the command `yarn build-icons` the typescript icons generated don't have the same formatting.I Don't know if it's come from my config or do we change something?


vindhya (changes requested):
Looks good - just one thing: The file, `src/icons/Menu.tsx` can be deleted.

ffffffelix:
I updated `yarn build-icons` to remove typescript icons `src/icons/*.tsx` before running to keep sync with SVG icons.

vindhya (approved):

PR #251, opened 2021-06-03
MOON-287: Table Component Structured View

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-287

## Description
- Created `ExpansionContainer` component to contain logic and styles for the structured view chevrons
- Updated `ListItem` component to take optional `typographyVariant` and `iconSize` props that will allow the user to control these aspects of the component. They still default to the prior values so it won't cause any changes to the existing `ListItem` implementations
- Updated `tableDataNested.tsx` sample data file


src/components/Table/Table.stories.jsx
```diff
@@ -387,8 +416,97 @@ export const Pagination = () => {
...
+                        depth={row.depth}
+                        getToggleRowExpandedProps={row.getToggleRowExpandedProps}
+                    >
+                        <ListItem
+                            iconSize="default"
+                            typographyVariant="body"
+                            iconStart={row.original.name.icon}
+                            label={row.values.name}
+                        />
+                    </ExpansionContainer>
```
ffffffelix, line 441: Can we embed this to the `TableCell` component and passing the needed props to it, to avoid exposing `ExpansionContainer` and `ListItem` and simplify the implementation of the structured view.

src/components/Table/table-cells/TableCell.scss
```diff
@@ -0,0 +1,26 @@
...
+    padding-right: var(--spacing-medium);
+
+    &:first-child {
+        padding-left: var(--spacing-medium);
+    }
+
+    .moonstone-TableCell_children {
+        display: -webkit-box;
+        -webkit-line-clamp: 1;
+        -webkit-box-orient: vertical;
```
ffffffelix, line 11: As we have autoprefixer now, we don't need to prefix our CSS properties. What about non-webkit browser?

src/components/Table/Table.stories.jsx
```diff
@@ -62,7 +63,23 @@ export const Basic = () => (
...
+            accessor: row => row.name.value,
+            Cell: cellInfo => {
+                const {row} = cellInfo;
+                return (
+                    <ListItem
+                        iconSize="default"
+                        typographyVariant="body"
+                        iconStart={row.original.name.icon}
+                        label={row.values.name}
+                    />
```
ffffffelix, line 78: As the `ListItem` component render a `li` we can't use it here because of the semantic.

src/components/Table/ExpansionContainer/ExpansionContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const paddingLeft = `${depth * 1.5}rem`;
+    const marginLeft = '16px';
+
+    if (canExpand) {
+        return (
+            <div
+                {...getToggleRowExpandedProps({style: {paddingLeft}})}
+                className={clsx('moonstone-ExpansionContainer', 'flexRow_nowrap', 'alignCenter', className)}
+            >
+                <span {...props}>
```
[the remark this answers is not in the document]
vindhya: I was initially thinking that anyone who wants to apply props to this component would want to apply it as close as possible to the actual expansion icon, but I think you're correct, putting it on the root element would be more consistent with what we're doing in the rest of Moonstone. I'm making the change :)

src/components/Table/Table.stories.jsx
vindhya: Unfortunately, because `TableCell` wraps the rendered value in the cell, it's difficult to render the chevron right/down icons in the correct place in relation to the content type icon. So, in this case, this `ExpansionContainer` that's rendered, gets wrapped by the `TableCell`. Maybe we can have a quick call about this so that I can explain it better.

src/components/Table/table-cells/TableCell.scss
vindhya: `line-clamp` requires it 🤷🏾‍♀️

ffffffelix:
Nice job 🙌

I have an issue with the structured view when I collapse/expand some nodes with long text, the column width changed. Can you confirm it will be solved with #252.

Refactor for next time: See if we can avoid having too many nested elements inside a `td`.

vindhya:
@ffffffelix Yes, that issue should be resolved when using any of `react-table`'s layout plugins which is what we did in #252 🙂

ffffffelix (approved):

PR #253, opened 2021-06-12
MOON-297: Create "kitchen sink" story examples with all Table component features

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-297

## Description

- Create a flat data story and a nested data story to show Table component with all features included
- Some cleanup to Table components (fixes some layout bugs)

### DIFF
--- a/src/components/IconTextIcon/IconTextIcon.scss
+++ b/src/components/IconTextIcon/IconTextIcon.scss
@@ -1,11 +1,11 @@
 .moonstone-IconTextIcon {
     width: 100%;
+}
 
-    .moonstone-IconTextIcon_iconStart {
-        margin-right: var(--spacing-nano);
-    }
+.moonstone-IconTextIcon_iconStart {
+    margin-right: var(--spacing-nano);
+}
 
-    .moonstone-IconTextIcon_iconEnd {
-        margin-left: var(--spacing-nano);
-    }
+.moonstone-IconTextIcon_iconEnd {
+    margin-left: var(--spacing-nano);
 }
--- a/src/components/IconTextIcon/IconTextIcon.scss
+++ b/src/components/IconTextIcon/IconTextIcon.scss
@@ -1,7 +1,11 @@
-.moonstone-IconTextIcon_iconStart {
-    margin-right: var(--spacing-nano);
-}
+.moonstone-IconTextIcon {
+    width: 100%;
 
-.moonstone-IconTextIcon_iconEnd {
-    margin-left: var(--spacing-nano);
+    .moonstone-IconTextIcon_iconStart {
+        margin-right: var(--spacing-nano);
+    }
+
+    .moonstone-IconTextIcon_iconEnd {
+        margin-left: var(--spacing-nano);
+    }
 }
--- a/src/components/IconTextIcon/IconTextIcon.tsx
+++ b/src/components/IconTextIcon/IconTextIcon.tsx
@@ -22,7 +22,7 @@ export const IconTextIcon: React.FC<IconTextIconProps> = ({
     return React.createElement(
         component,
         {
-            className: clsx('flexRow', 'alignCenter'),
+            className: clsx('moonstone-IconTextIcon', 'flexRow_nowrap', 'alignCenter'),
             ...props
         },
         (
@@ -34,7 +34,12 @@ export const IconTextIcon: React.FC<IconTextIconProps> = ({
                     />
                 )}
 
-                <Typography className="flexFluid" {...typoProps}>{children}</Typography>
+                <Typography
+                    className={clsx('flexFluid')}
+                    {...typoProps}
+                >
+                    {children}
+                </Typography>
 
                 {iconEnd && (
                     <iconEnd.type
--- a/src/components/IconTextIcon/IconTextIcon.tsx
+++ b/src/components/IconTextIcon/IconTextIcon.tsx
@@ -22,7 +22,7 @@ export const IconTextIcon: React.FC<IconTextIconProps> = ({
     return React.createElement(
         component,
         {
-            className: clsx('moonstone-IconTextIcon', 'flexRow_nowrap', 'alignCenter'),
+            className: clsx('moonstone-IconTextIcon', 'flexRow_nowrap', 'alignCenter', className),
             ...props
         },
         (
@@ -35,6 +35,7 @@ export const IconTextIcon: React.FC<IconTextIconProps> = ({
                 )}
 
                 <Typography
+                    isNowrap
                     className={clsx('flexFluid')}
                     {...typoProps}
                 >
--- a/src/components/Table/table-cells/TableBodyCell.tsx
+++ b/src/components/Table/table-cells/TableBodyCell.tsx
@@ -14,18 +14,16 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
     className,
     iconStart,
     iconEnd,
-    expandableColumnIndex,
+    isExpandableColumn,
     row,
     cell,
     children,
     ...props
 }) => {
-    // Fix this type definition of any
-    const isExpandableColumn = (row as any)?.cells[expandableColumnIndex].value === cell?.value;
     const leftMarginBuffer = 20; // px
     const leftMarginIndentDepth = row?.depth * 20; // px
 
-    const renderIconTextIcon = () => (
+    const renderCellContent = () => (
         <IconTextIcon
             component="div"
             iconStart={iconStart}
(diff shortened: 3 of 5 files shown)


src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -15,16 +14,67 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
...
+        // These are cells that are in the expandable row (canExpand) and it is the column in
+        // which the cells show the chevron icon to expand and collapse sub-rows (isExpandableColumn)
+        if (isExpandableColumn && row?.canExpand) {
+            return (
+                <TableCell
+                    {...
+                        (row?.canExpand && isExpandableColumn
+                            ? row?.getToggleRowExpandedProps({style: {marginLeft: `${leftMarginIndentDepth}px`}})
+                            : {}
+                        )
```
ffffffelix, line 48: As you already check this condition above with `if (isExpandableColumn && row?.canExpand)` this condition seems unnecessary.

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -15,16 +14,67 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
...
+    cell,
     children,
     ...props
 }) => {
-    const paddingLeft = `${depth * 0.75}rem`;
-    const marginLeft = spacings.spacingMedium;
+    // Fix this type definition of any
+    const isExpandableColumn = (row as any)?.cells[expandableColumnIndex].value === cell?.value;
+    const leftMarginBuffer = 20; // px
+    const leftMarginIndentDepth = row?.depth * 20; // px
```
ffffffelix, line 26: Do you think is it relevant to use `leftMarginBuffer` in place of the numeric value `20`.
```suggestion
    const leftMarginIndentDepth = row?.depth * leftMarginBuffer; // px
```

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -15,16 +14,67 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
...
+        // but do not have the chevron to expand/collapse rows underneath them.
+        // Also, a buffer of 20px is added so that they are aligned with the cells that do have
+        // the chevron icons for expand/collapse
+        if (isExpandableColumn && !row?.canExpand) {
+            return (
+                <TableCell style={{marginLeft: `${leftMarginIndentDepth + leftMarginBuffer}px`}}>
+                    {renderIconTextIcon()}
+                </TableCell>
+            );
+        }
```
ffffffelix, line 71: The cells without children at the root level are also affected and they are not aligned with the header anymore. I think we also need to check the depth to add the `leftMarginBuffer`.

```suggestion
        if (isExpandableColumn && !row?.canExpand && row?.depth > 1) {
            return (
                <TableCell style={{marginLeft: `${leftMarginIndentDepth + leftMarginBuffer}px`}}>
                    {renderIconTextIcon()}
                </TableCell>
            );
        }
```

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -15,16 +14,67 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
...
+
+        // These are just the normal cells in the other columns which don't display anything with
+        // relation to the row expansion feature
+        return <TableCell>{renderIconTextIcon()}</TableCell>;
```
ffffffelix, line 75: It's "funny" because in some cases we neither display icons nor a text in the cell, but a `Chip` for example. But I still don't find a better name

src/components/Table/table-cells/TableCell.types.ts
```diff
@@ -33,29 +34,20 @@ export type TableCellProps = {
...
     /**
-     * Whether this cell is in the first column of the table
+     * Row object returned by react-table instance
      */
-    canExpand?: boolean;
+    row?: UseExpandedRowProps<Row>;
 
     /**
-     * Whether the cell is currently expanded
+     * Cell object returned by react-table instace
```
ffffffelix, line 48: Just a typo
```suggestion
     * Cell object returned by react-table instance
```

src/components/IconTextIcon/IconTextIcon.scss
```diff
@@ -1,7 +1,11 @@
-.moonstone-IconTextIcon_iconStart {
-    margin-right: var(--spacing-nano);
-}
+.moonstone-IconTextIcon {
+    width: 100%;
```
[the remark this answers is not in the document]
vindhya: Without it, the text stretches past its container. I.e., the `text-overflow: ellipses` doesn't get displayed when using `IconTextIcon` in the Table.

src/components/IconTextIcon/IconTextIcon.tsx
```diff
@@ -34,7 +34,12 @@ export const IconTextIcon: React.FC<IconTextIconProps> = ({
                     />
                 )}
 
-                <Typography className="flexFluid" {...typoProps}>{children}</Typography>
+                <Typography
+                    className={clsx('flexFluid')}
+                    {...typoProps}
+                >
```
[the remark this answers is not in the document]
vindhya: I can add `isNowrap`, but it doesn't seem to do anything. Also as above, the `Typography` width isn't constrained to the cell without using `width: 100%` in the `IconTextIcon` component scss.

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -15,16 +14,67 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
...
+    expandableColumnIndex,
+    row,
+    cell,
     children,
     ...props
 }) => {
-    const paddingLeft = `${depth * 0.75}rem`;
-    const marginLeft = spacings.spacingMedium;
+    // Fix this type definition of any
+    const isExpandableColumn = (row as any)?.cells[expandableColumnIndex].value === cell?.value;
```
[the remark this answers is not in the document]
vindhya: `canExpand` is a property that comes from react-table which indicates whether the entire row can be expanded. However, we could move the logic outside the component to something like, `canColumnExpand`. Maybe that's simpler?

The check is going to happen either way: it's just whether it happens outside of the Moonstone component or within it. I could see how it makes sense to keep the Moonstone more simple and do the check outside the component and pass in the boolean via a prop.

src/components/IconTextIcon/IconTextIcon.tsx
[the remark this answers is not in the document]
vindhya: Ohh you mean set it as default for the `Typography` component inside `IconTextIcon` instead of setting it from outside `IconTextIcon` and passing it in....sure :) that works for me 👍🏾

src/components/Table/table-cells/TableBodyCell.tsx
vindhya: Hmm, not sure if that can be removed. It's the case where the cell is in the expandable column (so it needs proper indentation), but the row itself doesn't have any nested sub-rows.

The condition after that is for all the other table cells - the ones that aren't in an expandable column.

src/components/Table/table-cells/TableBodyCell.tsx
vindhya: The only reason I thought so was to have all the constant values easily located at the top of the component. But, I'm happy to go either way on this.

src/components/Table/table-cells/TableBodyCell.tsx
vindhya: I actually wanted cells without children at the root level affected by this as well because we need room to show the expansion chevron on the left of the content inside the cell. I think it would be confusing if the the content part of all the root level cells didn't line up. And, I don't think we should push them to the left because we can't rely on there being extra space on the left side of the Table (in the case that there is other column to the left of the expandable column).

src/components/Table/table-cells/TableBodyCell.tsx
vindhya: Maybe changing the name of that render function to something like `renderContent()` would be less funny?

AKarmanov (approved):
Looks like all changes are in and we should be god to go :)

src/components/Table/table-cells/TableBodyCell.tsx
ffffffelix: I understand

src/components/Table/table-cells/TableBodyCell.tsx
ffffffelix: I like the idea to change the name by something more generic.

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -38,7 +37,7 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
     const renderTableCell = () => {
         // These are cells that are in the expandable row (canExpand) and it is the column in
         // which the cells show the chevron icon to expand and collapse sub-rows (isExpandableColumn)
-        if (isExpandableColumn && row?.canExpand) {
+        if (isExpandableColumn && row?.canExpand && row?.depth > 1) {
```
vindhya, line 40: @ffffffelix I reverted this change because it was causing a bug in rendering the expansion chevrons. I don't think it was necessary, anyway.

src/components/Table/table-cells/TableBodyCell.tsx
vindhya: Removed the redundant conditional check.

src/components/Table/table-cells/TableBodyCell.tsx
vindhya: Changed the name of the function to `renderCellContent`

PR #255, opened 2021-06-25
MOON-306: Add icon ViewTree

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-306

## Description

Add icon ViewTree


vindhya (approved):

tdraier:
Just wondering, is this icon generic enough to be in moonstone or is it  only for jcontent ? I'm feeling there are a lot of icons in moonstone specific to some apps, which should not be there

PR #257, opened 2021-06-29
QA-13381 Close menu when onContextMenu is triggered on overlay

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-13381

## Description

Basically equating right click on the overlay with a click. In jcontent a user can right click on entries in the table, if a menu is open a right click on the overlay doesn't change selection, although for the user it seems like a different entry is selected. Alternatively, we can disable right click but I think it may leave the user confused.


AKarmanov:
@vindhya let me know if this makes sense or if you can see a better solution for this.

vindhya (approved):

PR #258, opened 2021-07-06
MOON-301: Table Sticky Header

vindhya:
## JIRA

https://jira.jahia.org/browse/MOON-301

## Description

- Allows the header of a table to be sticky with the use of the `isSticky` prop
- Use of the sticky header causes issues with the `useFlexLayout` plugin, so its usages in `Table.stories.jsx` have been commented out for now. Further investigation into adding the ability to specify column widths will be done with MOON-299

### DIFF
--- a/src/components/Table/Table.stories.jsx
+++ b/src/components/Table/Table.stories.jsx
@@ -5,7 +5,7 @@
 */
 
 import React, {useEffect, useState} from 'react';
-import {useExpanded, useFlexLayout, useRowSelect, useSortBy, useTable} from 'react-table';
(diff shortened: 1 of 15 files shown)


src/components/Table/Table.stories.jsx
```diff
@@ -598,8 +632,8 @@ export const KitchenSinkFlat = () => {
             disableSortRemove: true
         },
         useSortBy,
-        useRowSelect,
-        useFlexLayout
+        useRowSelect
+        // UseFlexLayout
```
gflores-jahia, line 636: do we want to just remove instead of commenting out?

gflores-jahia (approved):

PR #260, opened 2021-07-22
BACKLOG-16279: Add data-sel-role tags to table pagination component

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16279

## Description

- Added data-sel-roles to some of the table pagination elements
- Added unit tests


ffffffelix (approved):
As we want to keep moonstone as generic as possible, we avoid adding `data-sel-role` to moonstone because it's very specific to our product.
Usually, the best practice should be to provide props to add any attributes to the desired element on the implementation side. But as we want to add a `data-sel-role` to all "interactive" elements and because the pagination can only be used in a table context. I think it can be good for this use case.

PR #261, opened 2021-07-22
MOON-299: Allow to add a width to each cells

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-299

## Description

- The table layout is managed with an old-good CSS flexbox, so we can't use `useFlexbox` or other 
plugins layout from react-table.
- We have to set a width to the `TableCellBody` **and** `TableCellHeader`
- When no width is provided the cell takes all space available
- When no width is set at all, or cells are the same size

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -12,7 +12,7 @@ export const Table: React.FC<TableProps> = ({
 }) => React.createElement(
     component,
     {
-        className: clsx('moonstone-Table', className),
+        className: clsx('moonstone-Table', 'flexCol_nowrap', className),
         ...props
     },
     children
--- a/src/components/Table/table-cells/TableHeadCell.tsx
+++ b/src/components/Table/table-cells/TableHeadCell.tsx
@@ -8,6 +8,7 @@ import {TableCell} from './TableCell';
 
 export const TableHeadCell: React.FC<TableCellProps> = ({
     component = 'th',
+    width,
     textAlign = 'left',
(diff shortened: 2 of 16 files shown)


src/components/Table/Table.tsx
```diff
@@ -12,7 +12,7 @@ export const Table: React.FC<TableProps> = ({
 }) => React.createElement(
     component,
     {
-        className: clsx('moonstone-Table', className),
+        className: clsx('moonstone-Table', 'flexCol_nowrap', className),
```
gflores-jahia, line 15: just checking...I see this added but not seeing it defined anywhere...

src/components/Table/table-cells/TableHeadCell.tsx
```diff
@@ -16,20 +17,19 @@ export const TableHeadCell: React.FC<TableCellProps> = ({
...
+                {'flexFluid': typeof width === 'undefined'},
                 'textAlign' + capitalize(textAlign),
                 'moonstone-verticalAlign' + capitalize(verticalAlign),
                 className
             )}
             component={component}
             weight="bold"
             variant="body"
-            {...props}
+            style={{width, ...props.style}}
```
gflores-jahia, line 32: might need to reverse this if width needs to override what's in props.style

ffffffelix: Nice catch. Yes, you're right I reversed the order

gflores-jahia (approved):

PR #263, opened 2021-07-29
MOON-305: POC for treeview in dropdown

tdraier:
## JIRA

https://jira.jahia.org/browse/MOON-305

## Description

This is POC - work in progress

### DIFF
--- a/src/components/Dropdown/Dropdown.scss
+++ b/src/components/Dropdown/Dropdown.scss
@@ -17,6 +17,18 @@ $border-width: 1px;
     margin-right: var(--spacing-nano);
 }
 
+.moonstone-disabled {
+    & > .moonstone-treeView_item {
+        @include is-disabled;
+
+        pointer-events: none;
+
+        &:hover {
+            pointer-events: none;
+        }
+    }
+}
+
 .moonstone-dropdown {
     padding-right: var(--spacing-small);
     padding-left: var(--spacing-small);
--- a/src/components/TreeView/ControlledTreeView.tsx
+++ b/src/components/TreeView/ControlledTreeView.tsx
@@ -7,6 +7,7 @@ import Loading from '~/icons/Loading';
 import ChevronDown from '~/icons/ChevronDown';
 import ChevronRight from '~/icons/ChevronRight';
 import {Typography} from '~/components/Typography';
+import {isDisabled} from "@testing-library/user-event/dist/utils";
 
 export const ControlledTreeView: React.FC<ControlledTreeViewProps> = ({
     data,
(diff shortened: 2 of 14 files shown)


src/components/Dropdown/Dropdown.scss
```diff
@@ -17,6 +17,18 @@ $border-width: 1px;
     margin-right: var(--spacing-nano);
 }
 
+.moonstone-disabled {
```
ffffffelix, line 20: It seems we can't disable an item with the `TreeViewMenu`.
- The prop `isDisabled` is not defined in the type`TreeViewData`
- Please, can you provide an example in storybook

tdraier: - isDisabled now added
- there's already an example in storybook ( playground / tree)

src/components/TreeView/ControlledTreeView.tsx
```diff
@@ -124,7 +126,7 @@ export const ControlledTreeView: React.FC<ControlledTreeViewProps> = ({
                             {/* TreeViewItem */}
                             <div
                                 className={clsx('flexRow_nowrap', 'alignCenter', 'flexFluid', 'moonstone-treeView_itemLabel', node.className)}
-                                onClick={handleNodeClick}
+                                onClick={!node.isDisabled && handleNodeClick}
```
ffffffelix, line 129: I have this warning
> If you used to conditionally omit it with onClick={condition && value}, pass onClick={condition ? value : undefined} instead.

src/components/TreeView/ControlledTreeView.tsx
```diff
@@ -7,6 +7,7 @@ import Loading from '~/icons/Loading';
 import ChevronDown from '~/icons/ChevronDown';
 import ChevronRight from '~/icons/ChevronRight';
 import {Typography} from '~/components/Typography';
+import {isDisabled} from "@testing-library/user-event/dist/utils";
```
ffffffelix, line 10: Why do you import something from `testing-library` into the component?

ffffffelix (commented):
I'm wondering if we should disable children when a tree node is disabled.

src/components/TreeView/ControlledTreeView.tsx
tdraier: no idea :-)

tdraier:
> I'm wondering if we should disable children when a tree node is disabled.

I think it's up to the component user to decide if he wants to disable all children, or even do not display the children at all. By just disabling the single tree node we leave the possibility to do what we want

ffffffelix:
> > I'm wondering if we should disable children when a tree node is disabled.
> 
> I think it's up to the component user to decide if he wants to disable all children, or even do not display the children at all. By just disabling the single tree node we leave the possibility to do what we want

If we let the component user decide, we may have different behaviors for the same component and so have inconsistency in our interface. It's the role of the design system to prevent some behaviors to ensure a homogeneous experience.
We will decide later which behavior we want for this use case.

ffffffelix (approved):

PR #269, opened 2021-09-07
Backlog 16359 scroll table cells

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16359

## Description

Allow cell to scroll when it contains overflowing content

### DIFF
--- a/src/components/Table/table-cells/TableBodyCell.tsx
+++ b/src/components/Table/table-cells/TableBodyCell.tsx
@@ -59,7 +59,7 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
         // the chevron icons for expand/collapse
         if (isExpandableColumn && !row?.canExpand) {
             return (
-                <TableCell style={{marginLeft: `${leftMarginIndentDepth + leftmarginbuffer}px`}}>
+                <TableCell style={{marginLeft: `${leftMarginIndentDepth + leftMarginBuffer}px`}}>
                     {renderCellContent()}
                 </TableCell>
             );
--- a/src/components/Table/table-cells/TableBodyCell.tsx
+++ b/src/components/Table/table-cells/TableBodyCell.tsx
@@ -20,15 +20,17 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
     row,
     cell,
     children,
+    isScrollable,
     ...props
 }) => {
     const leftMarginBuffer = 20; // px
     const leftMarginIndentDepth = row?.depth * 20; // px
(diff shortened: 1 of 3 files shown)


src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -27,7 +28,7 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
     // TODO: When no iconStart is set we should just render children to avoid too nested HTML elements
     // TODO: Add iconEnd somewhere or remove it from the props, it seems we never use it
     const renderCellContent = () => (
-        <IconTextIcon component="div" iconStart={iconStart}>
+        <IconTextIcon component="div" iconStart={iconStart} typographyProps={{className: clsx('moonstone-tableCellContent', 'flexFluid')}}>
```
ffffffelix, line 31: I think it's better to add a prop to manage if the cell is scrollable or not. The default behavior must still be to truncate the text when it's too long.
In that case, we can rename the CSS class by something more descriptive like `moonstone-tableBodyCell_scrollable`.

src/components/Table/table-cells/TableCell.scss
```diff
@@ -1,5 +1,23 @@
 @import '../../../utils/index.scss';
 
+.moonstone-tableCellComponent {
+    overflow-y: auto;
+
+    &::-webkit-scrollbar {
+        height: 2px;
```
ffffffelix, line 7: As the height only applies to WebKit browser the behavior is a little weird with other browsers like Firefox. I don't know how and if we can manage that ...

AKarmanov: I increased it to 4px actually, we can have it it doesn't distort anything in ff. I also added a similar prop for ff but it is still quite wide, covers content although it is semi transparent and you can still see a thing or two.

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -70,6 +71,7 @@ export const TableBodyCell: React.FC<TableCellProps> = ({
     return (
         <Typography
             className={clsx(
+                'moonstone-tableCellComponent',
```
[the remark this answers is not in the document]
AKarmanov: 👍

src/components/Table/table-cells/TableBodyCell.tsx
AKarmanov: 👍

ffffffelix (approved):

PR #278, opened 2021-11-30
MOON-321: Add Controlled/Uncontrolled Input

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-321

## Description

- Add Controlled/Uncontrolled Input
- Use a Button to clear the field
- Update tests and rewrite them with testing-library

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -1,88 +1,14 @@
-import React, {useEffect, useRef} from 'react';
-import clsx from 'clsx';
-import {Cancel, Search} from '~/icons';
-import './Input.scss';
-import {InputProps, InputSizes, InputVariants} from './Input.types';
+import React from 'react';
+import {InputProps} from './Input.types';
+import {UncontrolledInput} from './UncontrolledInput';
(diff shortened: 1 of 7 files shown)


src/components/Input/Input.tsx
```diff
@@ -1,88 +1,14 @@
...
-        {'moonstone-size_big': size === InputSizes.Big},
-        {'moonstone-disabled': isDisabled},
-        className
-    );
-    const inputFilled = value !== '';
-    const inputEmpty = value === '';
-    const searchRef = useRef(null);
+export const Input: React.FC<InputProps> = ({defaultValue, value, ...props}) => {
+    if (typeof value === 'undefined') {
+        return <UncontrolledInput defaultValue={defaultValue} {...props}/>;
```
tdraier, line 8: no need to extract defaultValue :
```suggestion
export const Input: React.FC<InputProps> = ({value, ...props}) => {
    if (typeof value === 'undefined') {
        return <UncontrolledInput {...props}/>;
```

tdraier (approved):

PR #283, opened 2021-12-15
MOON-323: Fix disabled and read-only checkbox state

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-323

## Description

- Fix disabled and read-only checkbox state
- Add tests


tdraier (commented):
isReadOnly should be added into CheckboxProps

tdraier (approved):

ffffffelix:
> isReadOnly should be added into CheckboxProps

The prop `isReadOnly` is already define

https://github.com/Jahia/moonstone/blob/04232d06b8071dd4978488f61c3265e72e6e4a66/src/components/Checkbox/Checkbox.types.ts#L50-L53

PR #284, opened 2021-12-15
MOON-317: Add the ability to set a color to icons and add more icons

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-317

### Icons
- Added icons: `Page`, `Archive`, `Unarchive`, `ZoomIn`, `ZoomOut`, `FilePDF`, `FileText`, `FileSound`, `FileVideo`, `FileCode`, `FileFont`, `FileDoc`, `FileExcel`, `FilePowerPoint`
- Updated icons: `FileCompressed`, `FileImage`, `File`
- Removed icon `Pdf`
- Add the ability to set a color to icons with the prop `color`. The accepted colors are `blue`, `red`, `yellow`, `green`, `purple`, and `gray`.

### Fix
- Add the ability to add `ClassName` to the element  used in `iconStart` and `iconEnd` from `IconTextIcon`
- When a TableRow is highlighted the icon is always with `color-light` even if the icon is colored.

### Project structure
Now all elements relative to icons are in `src/icons`
- `assets` contains SVG icons
- `components` contains icons components built with svgr
- `utils` contains `toIconComponents`

This new structure allows us to simplify our icon build process by removing the pre/post build and the `rename.js` script.
To be consistent with other scripts I renamed `build-icons` to `build:icons`


tdraier:
It could be interesting to have a default color in some icons (powerpoint, word, ...), so that they display in "color" when no color is specified. 
Note that it will not be possible (actually, quite hard and mainly inconsistent) to set the color depending on the file type from jcontent - we can only choose an icon component based on the type.

tdraier:
Actually it is possible with the current code but not really scalable and consistent (current code is not even the same for grid and table view, and will need to be refactored anyway)

src/components/Checkbox/Checkbox.spec.js
```diff
@@ -37,7 +37,7 @@ describe('Checkbox', () => {
     });
 
     it('should have mixed state when specified with the isIndeterminate prop', () => {
-        render(<Checkbox checked="mixed" aria-label="checkbox"/>);
+        render(<Checkbox defaultChecked="mixed" aria-label="checkbox"/>);
```
tdraier, line 40: I don't know why this change appears here, but defaultChecked is anyway a boolean and cannot take "mixed"

src/components/Table/TableRow/TableRow.scss
```diff
@@ -9,36 +9,40 @@ $row-multiple-line-height: 64px;
     .moonstone-TableBody &:hover {
```
tdraier, line 9: Also don't see the relation with icons here ?

src/components/Checkbox/Checkbox.spec.js
ffffffelix: You're right, maybe it was not the right place to fix that, but when I run the test I get this error:

> Warning: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.

So the solution will be to add an `onChange` function?

ffffffelix: I kept `checked=mixed` and added `onChange` to remove the warning

src/components/Table/TableRow/TableRow.scss
ffffffelix: Finally, I kept the nested classes and just added this scss snippet.

tdraier (approved):

PR #295, opened 2022-06-09
Fixed header separator

tdraier:


src/components/Header/Header.tsx
```diff
@@ -51,7 +51,7 @@ export const Header: React.FC<HeaderProps> = ({
                 <div className={clsx('flexRow_between', 'alignCenter', 'moonstone-header_information')}>
                     <div className={clsx('flexRow_nowrap', 'alignCenter', 'flexFluid', 'moonstone-header_informationLeft')}>
                         { breadcrumb }
-                        { breadcrumb && contentType && <Separator variant="vertical" spacing="medium"/> }
+                        { <Separator variant="vertical" spacing="medium" invisible="firstOrLastChild"/> }
```
ffffffelix, line 54: I'm wondering if it should be the default behavior of the component.
When the separator is the first or the last child we don't need to display it, instead of manage it by a prop.

ffffffelix (approved):

PR #298, opened 2022-07-04
TECH-568 migrate moonstone to github action

Janin-Michel-Mathias:


.github/workflows/on-code-change.yml
```diff
@@ -0,0 +1,31 @@
+name: On Code Change
```
ffffffelix, line 1: As the job is triggered on `push` maybe we can rename this workflow to fit what it does ?

.github/workflows/on-code-change.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    needs: lint-tests
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        shell: bash
+        run: yarn install --frozen-lockfile
+      - name: Run Build
+        shell: bash
+        run: yarn build
```
ffffffelix, line 30: The build process must be only triggered when we push on the `master` branch.
Can you split it to an another workflow for the build process ?

ffffffelix:
Some questions: 
- Should we remove the circleCI jobs as we replaced it with Github Actions?
- Do we need to migrate the release job from CircleCI to Github Actions?

Maybe we also can remove husky jobs to avoid to run lint and test jobs twice, on the local machine and on the server throw Github Actions.

ffffffelix (changes requested):

.github/workflows/on-code-change.yml
Janin-Michel-Mathias: It's triggered on push only because I didn't finish the migration and I test it each time I change something

.github/workflows/on-merge.yml
```diff
@@ -0,0 +1,62 @@
...
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v2
+      - uses: webfactory/ssh-agent@v0.5.4
+        with:
+          ssh-private-key: |
+            ${{ secrets.PERF_BASTION_PRIVATE_KEY }}
+            ${{ secrets.JAHIA_BASTION_PRIVATE_KEY }}
+      - uses: actions/download-artifact@v3
```
Fgerthoffert, line 30: @Janin-Michel-Mathias What is this for ? You're downloading an artifact called build-artifacts, but I don't see it uploaded anywhere else on this repo.

.github/workflows/on-merge.yml
```diff
@@ -0,0 +1,62 @@
+name: On merge
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
```
Fgerthoffert, line 9: Since build is not used by any job other than release-publish in this workflow, I'd be tempted to say that you can actually merge it with release-publish.

This will also address my other comment about the `actions/download-artifact`.

Fgerthoffert (changes requested):

ffffffelix (approved):

.github/workflows/on-merge.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install dependencies
+        shell: bash
+        run: yarn install --frozen-lockfile
+      - name: Run Build
+        shell: bash
+        run: yarn build
+      - uses: webfactory/ssh-agent@v0.5.4
+        with:
+          ssh-private-key: |
+            ${{ secrets.PERF_BASTION_PRIVATE_KEY }}
```
Fgerthoffert, line 23: I actually believe these keys are not the correct ones.

Looking at the commands below, the SSH key seems to be used to allow changes created by the changelog to be pushed back to the repo, so the key is most likely this one: `GH_SSH_PRIVATE_KEY_JAHIACI`

Fgerthoffert (changes requested):

PR #299, opened 2022-07-13
MOON-336: Add searchContext for Input

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-336

## Description
- Create an internal component `BaseInput` to manage all Inputs
- The `variant` prop of the Input component becomes deprecated. To keep backward compatibility the component`SearchInput`is used when `variant` is set to `search`, but I added a console.warn to notify the deprecation.
- Create a new component `SearchInput` based on `BaseInput`.
- Create a new component `searchContextInput` to manage the search context, based on `baseInput`
- Share `dropdownData` structures into `src/data` to avoid recreating data structures for tests and stories
- Rewrite input stories in typescript with CSF storybook format


tdraier:
The relationship between variant=search, size and searchContext is not very clear and each of them has impact on the others. Searchcontext does not appear if variant is not search, and size will be ignored if searchContext is specified, so some combination will give unexpected results and will be confusing. It would be more clear to remove interactions between the props and let the user choose what he wants to use (even if a small input with search context and custom icon is not supposed to be used)
Actually I don't really understand why we have a "search" variant (variant is more to define the appearance of the button, not it's usage/content) - it looks more like a "predefined configuration" than a variant.

ffffffelix:
For the Input component, the `variant` prop has mostly the same usage as the HTML attribute `type`. But, we can't use `type` as browsers have different implementations, and it is difficult to override them. So the idea was to have something to define specific behaviors for some input type as search, date, password, number, color etc...

The variant search sets the search icon and the clear button at the end automatically.

In that case:
-  Having another variant/type makes no sense when we use `SearchContext`.
- A dropdown inside the Input doesn't fit well when we use the small size.

As the appearance must reflect the usage, the split between both is very blurry. In my mind, the Design System must be the gatekeeper of consistency. So if we can prevent unwanted props combinations and prevent errors during the implementation by presetting some, I think we should do it. Plus, by "forcing" some props into the Design System, it will be easier to manage when we want to change the behaviors we just have to do it in one place.

But you're right. Maybe the prop `variant` is badly named for this purpose.

ffffffelix:
I created a new component `SearchContextInput` to manage search with a context. For now, I mostly duplicate the code from the Input component.

tdraier:
I was just thinking of a SearchInput which uses Input and force some props. But that's only an idea, I'm not sure it will solve the searchContext part

ffffffelix:
I think it's a good idea. So I created a `BaseInput` only for internal purposes to manage all inputs and deprecate the prop `variant` for input. So now we have:
- `Input` for basic purposes
- `SearchInput` for search. It automatically sets the right `icon` and `role`
- `SearchContextInput`to manage context with search set also the right `size` and `role`

tdraier (approved):

PR #313, opened 2022-09-29
MOON-314: Use fixed position

tdraier:
## JIRA

https://jira.jahia.org/browse/MOON-314


ffffffelix:
Setting the position to `absolute` for the dropdown menu has been done on purpose to fix issues ([MOON-223](https://jira.jahia.org/browse/MOON-223), [TECH-249](https://jira.jahia.org/browse/TECH-249), [MOON-80](https://jira.jahia.org/browse/MOON-80)). We have to be careful not to reintroduce bugs.

tdraier:
I'm forcing the positioning to absolute in case there is a parent which has a transform property. In other cases, keep fixed

ffffffelix (approved):

PR #332, opened 2022-11-23
MOON-353: support description on values in dropdown

jahia-carp:
https://jira.jahia.org/browse/MOON-353


src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -10,6 +10,7 @@ import {Love} from '~/icons';
 import {dropdownData, dropdownDataGrouped, dropdownDataImages, dropdownDataTree} from '~/data';
 import IconWrapper from '~/__storybook__/IconWrapper';
 import {iconsName} from '~/__storybook__/utils';
+import {Folder, Lock} from '../../icons';
```
ffffffelix, line 13: To simplify the path you can use the alias `~` for `src`
```suggestion
import {Folder, Lock} from '~/icons';
```

src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -229,4 +230,106 @@ storiesOf('Components/Dropdown', module)
...
+        return (
+            <div style={{transform: 'scale(1)', height: '100vh', padding: '90px'}}>
+                <Dropdown
+                    icon={<Love/>}
+                    label={currentOption.label}
+                    value={currentOption.value}
+                    size={select('Size', DropdownSizes, DropdownSizes.Small)}
+                    isDisabled={boolean('Disabled', false)}
+                    maxWidth={text('Max width', '320px')}
+                    data={[
```
ffffffelix, line 251: A suggestion is to simplify the story and mutualize the data (even if I don't know if we will reuse it). You can create a new data file in `src/data`.
You have a data file per data object.

src/components/ListItem/ListItem.tsx
```diff
@@ -30,44 +31,97 @@ export const ListItem: React.FC<ListItemProps> = ({
...
             {...props}
         >
-            {
-                iconStart && !image && (
-                    <div className="moonstone-listItem_iconStart">
-                        <iconStart.type {...iconStart.props} size={iconSize}/>
-                    </div>
-                )
-            }
+            <ConditionalWrapper
```
ffffffelix, line 34: I'm wondering if it will not be easier to maintain/read by adding a condition based on `image` to return two possible renders because we use `image` for almost all conditions.
I'm afraid to add a lot of complexity to this component.

src/components/ListItem/ListItem.types.ts
```diff
@@ -13,6 +13,11 @@ export type ListItemProps = {
      */
     label: React.ReactNode;
 
+    /**
+     * ListItem description
```
ffffffelix, line 17: Please, use the same description for the same props
```suggestion
     * Optional description to display to describe the item
```

ffffffelix (changes requested):

src/components/ListItem/ListItem.tsx
tdraier: I don't get the point of this ConditionalWrapper at all .. ? it complexifies a lot and do not bring anything ?

jahia-carp: Fixed. 
I went to a lot of trouble for nothing, trying to avoid adding an unnecessary div if there was no description. But the conditionWrapper does make reading the code more complex than it should be. 
Thanks :-)

src/components/Dropdown/Dropdown.stories.jsx
jahia-carp: Done

src/components/Dropdown/Dropdown.stories.jsx
jahia-carp: Done

src/components/ListItem/ListItem.types.ts
jahia-carp: Done

src/components/ListItem/ListItem.tsx
```diff
@@ -30,42 +30,90 @@ export const ListItem: React.FC<ListItemProps> = ({
...
+                            'alignCenter'
+                        )}
+                        >
+                            {image}
+                        </figure>
+                    )
+                }
+
+                {
+                    image && !iconStart && (
```
tdraier, line 57: this test is similar to the previous one, blocks can be merged

src/components/ListItem/ListItem.tsx
```diff
@@ -30,42 +30,90 @@ export const ListItem: React.FC<ListItemProps> = ({
...
+                        <div className="flexCol">
+                            <Typography
+                                isNowrap
+                                isHtml={isHtml}
+                                className={clsx('flexFluid')}
+                                variant={typographyVariant}
+                                component="span"
+                            >
+                                {label}
```
tdraier, line 66: I don't clearly understand the conditions to display the label of the item - it seems that no label is displayed when `image && iconStart` , is it ok ?? It was inconditionnaly displayed before

ffffffelix: I simplify the component it should be clearer

tdraier (approved):

ffffffelix (approved):

PR #342, opened 2022-12-09
BACKLOG-20485 Get rid of unnecessary vars, add drop targets on lists,…

AKarmanov:
… update props

## JIRA

https://jira.jahia.org/browse/BACKLOG-20485

## Description

Get rid of unnecessary vars, add drop targets on lists, protect from dropping something that is not our list item, various improvements.


AKarmanov:
Wrong branch

PR #348, opened 2023-01-26
MOON-333 : Multiple selection in dropdown

tdraier:
## JIRA

https://jira.jahia.org/browse/MOON-333


src/components/Input/SearchContextInput/SearchContextInput.stories.tsx
```diff
@@ -46,15 +46,15 @@ const searchContextData: DropdownDataOptions[] = [
         iconStart: <Collections/>
     },
     {
-        label: 'Site',
+        label: 'Site Site|Site|Site|Site|Site|Site|Site| !!!',
```
ffffffelix, line 49:
```suggestion
        label: 'Site',
```

src/components/Menu/Menu.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {useState, useEffect} from 'react';
+import React, {useState, useEffect, ReactElement} from 'react';
```
ffffffelix, line 1: As you use `React.ReactElement`, you don't need this import.
```suggestion
import React, {useState, useEffect} from 'react';
```

src/components/Dropdown/Dropdown.scss
```diff
@@ -72,16 +72,6 @@ $border-width: 1px;
...
-
-    &:active {
-        color: var(--color-light);
-
-        background-color: var(--color-accent);
-    }
-
-    &.moonstone-opened {
-        @extend %dropdown-opened;
-    }
```
ffffffelix, line 84: It seems we need to keep this class to keep the blue border while the dropdown menu is opened. Currently when we use a multiple dropdown and I select an item the border disappears.

tdraier: dropdown-opened was only about putting a blue background : 
```
%dropdown-opened {
    color: var(--color-light);

    background-color: var(--color-accent);
}
```

The border is coming from somewhere else ?

ffffffelix: Currently, we only display the border for the `outlined` variant and not for the `ghost` variant; and I think we should have the same style for both variants.

As we replaced the blue background with a blue border when the dropdown menu is opened, I suggest using the class `.dropdown-opened`  to manage the style when the menu is opened.

ffffffelix (approved):

PR #364, opened 2023-02-10
QA-14659 Remove item from right-list only via icon click

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-14659

## Description

Remove item from right-list only via icon click. We still allow clicks on left-list items.


AKarmanov:
Had to fix a test

gflores-jahia (approved):

PR #380, opened 2023-06-14
MOON-373 Add z-index tokens

AKarmanov:
## JIRA

https://jira.jahia.org/browse/MOON-373

## Description

Add z-index tokens


ffffffelix (commented):
I'm wondering if we don't have too many values.
Maybe we can remove some intermediate values like `8500`, `9500`, `10500` and pick `8000` to the menu overlay (I hope it does not break something somewhere).

ffffffelix (approved):

PR #381, opened 2023-06-30
QA-14724 Shrink resize handle

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-14724

## Description

Shrink resize handle

### DIFF
--- a/src/components/ResizableBox/ResizableBox.scss
+++ b/src/components/ResizableBox/ResizableBox.scss
@@ -1,11 +1,13 @@
 .moonstone-resizable_handle {
+    /* stylelint-disable-next-line declaration-no-important */
+    right: -3px !important;
+
     display: flex;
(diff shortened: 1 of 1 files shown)


src/components/ResizableBox/ResizableBox.scss
```diff
@@ -4,7 +4,7 @@
     align-items: center;
 
     /* stylelint-disable-next-line declaration-no-important */
-    width: 18px !important;
+    width: 3px !important;
```
ffffffelix, line 7: You also have to add a `right` property to stick the handle to the SecondaryNavigation otherwise, there is a little space between both. It will become easier to grab it.
```suggestion
    width: 3px !important;
    right: -3px !important;
```

ffffffelix (changes requested):

PR #409, opened 2023-12-07
MOON-407: Add new component Pill

ffffffelix:
## JIRA

https://jira.jahia.org/browse/MOON-407

## Description

I added a new component ~~`ListItemChip`~~  `Pill` to display a Chip in place of `iconEnd` in Dropdown menu.
Actually, the purpose of this component is to display the language code in our Dropdown language switcher.

I hesitate to add this new style as a  variant of the `Chip` instead of creating a new component, but as we don't have the same props (no `icon`) and some specific (reversed style) I created a new component.

I don't really like the name of the component. I think it should be generic if we want to display something else than language code and should describe where we expect to use it. As a reviewer feel free to propose a better name.

- I created ~~`ListItemChip`~~  `Pill` and added some tests
- I replaced the icon sizes of the Dropdown to allow ~~`ListItemChip`~~ `Pill` from `width` to `min-width`
- I added an example in storybook in `Dropdown` to show what can look like the new language switcher


ffffffelix:
Finally, I renamed it `Pill`

gflores-jahia (approved):

PR #411, opened 2023-12-19
BACKLOG-21345: Add data attribute for tree item depth

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21345

## Description

Ran into an issue trying to select tree item for a given tree depth using style attribute.

Add a dedicated data attribute for specifying tree item depth to simplify css attribute selection.

Related PR context: https://github.com/Jahia/jcontent/pull/1129#issuecomment-1863466587


ffffffelix (commented):
I propose to use the attribute `aria-level` to add the depth information on the element ([MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-level))
What do you think?

ffffffelix (approved):

PR #416, opened 2024-05-07
BACKLOG-22393: Allow variable image sizes for Dropdown

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22393

## Description

- Remove fixed categorical sizes and let image size dictate the height of the row and width of image, up to a max width/height.
- Fixed `.moonstone-listItem-image_undefined` classes and default imageSize to small if not specified

### DIFF
--- a/src/components/ListItem/ListItem.scss
+++ b/src/components/ListItem/ListItem.scss
@@ -10,27 +10,25 @@ $big-image-size: 96px;
 
     list-style: none;
 
-    .moonstone-listItem-image_small {
-        flex: 0 0 $small-image-size;
-        height: $small-image-size;
+    .moonstone-listItem-image {
         margin-right: var(--spacing-small);
 
         img {
-            max-width: $small-image-size;
-            max-height: 100%;
             margin: auto;
         }
     }
 
-    .moonstone-listItem-image_big {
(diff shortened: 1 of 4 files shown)


src/components/ListItem/ListItem.scss
```diff
@@ -10,27 +10,25 @@ $big-image-size: 96px;
...
             margin: auto;
         }
     }
 
-    .moonstone-listItem-image_big {
-        flex: 0 0 $big-image-size;
-        height: $big-image-size;
-        margin-right: var(--spacing-small);
+    .moonstone-listItem-image_small {
+        img {
```
ffffffelix, line 22: I think the `img` tag is too specific here. We suppose that an `img` tag will pass through our `image` prop, but a [picture](https://developer.mozilla.org/fr/docs/Web/HTML/Element/picture) tag can also be used to display an image in HTML.

So, we should use `.moonstone-listItem-image_small > *` to select all direct children and not specify any element, as we don't really know what we will receive. (I know this it what we will have before ...)

Why do you think?

ffffffelix (commented):
Your test seems to fail. Can you take a look?

PR #442, opened 2024-10-15
MOON-438: Migrate tests of SecondaryNav

Eevolee:
## JIRA

https://jira.jahia.org/browse/MOON-438

## Description

Migrate tests of SecondaryNav from react component-test-utils to react testing-library

### DIFF
--- a/src/components/SecondaryNav/SecondaryNav.spec.js
+++ b/src/components/SecondaryNav/SecondaryNav.spec.js
@@ -18,14 +18,15 @@ describe('SecondaryNav', () => {
         expect(screen.getByTestId('secondary-nav')).toHaveAttribute('data-custom', 'extra');
     });
 
-    it('should have .secondaryNav_hidden when the menu is hidden', () => {
+    it('should not be expanded when the menu is hidden', () => {
         render(<SecondaryNav data-testid="secondary-nav" isDefaultVisible={false}>content here</SecondaryNav>);
-        expect(screen.getByTestId('secondary-nav')).toHaveClass('moonstone-secondaryNav_hidden');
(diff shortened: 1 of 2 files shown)


src/components/SecondaryNav/SecondaryNav.spec.js
```diff
@@ -1,77 +1,60 @@
...
+        render(<SecondaryNav data-testid="secondary-nav" isDefaultVisible={false}>content here</SecondaryNav>);
+        expect(screen.getByTestId('secondary-nav')).toHaveClass('moonstone-secondaryNav_hidden');
     });
 
     it('should not have .secondaryNav_hidden when the menu is visible', () => {
-        const wrapper = shallow(<SecondaryNav header="my header">hello</SecondaryNav>);
-
-        expect(wrapper.html()).not.toContain('secondaryNav_hidden');
+        render(<SecondaryNav data-testid="secondary-nav">content here</SecondaryNav>);
+        expect(screen.getByTestId('secondary-nav')).not.toHaveClass('moonstone-secondaryNav_hidden');
```
ffffffelix, line 28: We use a lot of classes for these tests. It does not respect the testing-libary [guideline](https://testing-library.com/docs/guiding-principles).
I suggest adding an `aria-expanded` attribute to the component and `aria-controls` to the button that toggles visibility.
- We will improve the semantic
- We will improve the accessible
- CSS class can change, not aria attribute

Can you change the tests to avoid the presence of CSS classes and use aria-* attributes when it possible ?

ffffffelix (changes requested):

ffffffelix (approved):

PR #460, opened 2024-11-27
MOON-338: Add component CardSelector

Eevolee:
## JIRA
https://jira.jahia.org/browse/MOON-338

## Description
Add components `CardSelector` & `EmptyCardSelector`
Component `CardSelector` has dragging icon enabled with `isDraggable` prop but is not currently handling dragging


gflores-jahia (dismissed):
For reference this is the jcontent-specific implementation from the old design system https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/DesignSystem/ReferenceCard/ReferenceCard.jsx

It looks like this has support for most of the information that's currently included in it, with a small comment regarding `EmptyCardSelector`.

src/components/CardSelector/EmptyCardSelector/EmptyCardSelector.tsx
```diff
@@ -0,0 +1,48 @@
...
+                'flexCol_center',
+                'flexFluid',
+                'alignCenter',
+                className
+            )}
+            aria-disabled={isDisabled || isReadOnly}
+            onClick={e => handleOnClick(e)}
+            {...props}
+        >
+            {label && <Typography className={clsx('moonstone-emptyCardSelector_label')}>{label}</Typography>}
```
gflores-jahia, line 41: From jcontent usage, looks like we make use of both icon and label for the empty card display. Is it possible to have them here as well? https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/DesignSystem/ReferenceCard/ReferenceCard.jsx#L195-L200

ffffffelix: Yes, we will add the ability to manage the icon for the EmptyCardSelector.

src/components/CardSelector/CardSelector.tsx
```diff
@@ -0,0 +1,141 @@
...
+        (e.currentTarget as HTMLElement).blur();
+    };
+
+    if (hasError) {
+        return (
+            <div
+                ref={ref}
+                id={id}
+                className={clsx('moonstone-cardSelector_error', 'flexRow_center', 'alignCenter')}
+                {...props}
```
ffffffelix, line 52: We also must handle `disabled` and `readonly` when the `CardSelector` is on error

ffffffelix, line 52: Add `onClick={e => handleOnClick(e)}` when the `CardSelector` is on error

src/components/CardSelector/EmptyCardSelector/EmptyCardSelector.types.ts
```diff
@@ -0,0 +1,39 @@
+import React, {HTMLAttributes} from 'react';
+
+export type EmptyCardSelectorProps = {
+    /**
+    * Id of the component
+    */
+    id?: string;
```
ffffffelix, line 7: As we have a required `id` I think we also request for a required `id` here.

ffffffelix (changes requested):
I think we forgot to add the import of the components in `src/components/index.ts`

ffffffelix (approved):

PR #461, opened 2024-12-05
MOON-246: Add Field component

Eevolee:
## JIRA

https://jira.jahia.org/browse/MOON-246

## Description

Add Field, FieldBoolean & FieldSelector components


gflores-jahia:
Just for reference as follow-up from our discussion, here's a breakdown of the different component layers for rendering fields in jcontent. Starting point is the `<Field>` in jcontent https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.jsx

- `<Field>` (container): takes care of behaviour actions (onChange, onBlur) per field and error handling through formik which calls renderField
- `renderField()` (in Field): renders input labels, actions, chips badges, etc and determines whether to render 
`<MultipleField>` and `<SingleField>`
  - This seems to be the equivalent of the new moonstone Field. Could be good to compare these two, specifically around rendering of errors.
Multiple:
- `<MultipleField>`: coordinates changes between fields (onChange, onFieldRemove, onFieldAdd, onReorder) and renders each field using `OrderableValue`
- `<OrderableValue>`: handles drag and drop events (using react-dnd) and rendering of each multiple field (wrapped with `<FastField>` (from formik)
Single:
- `<SingleField>`: component that wraps component with `<FastField>` (from formik)

ffffffelix:
> Just for reference as follow-up from our discussion, here's a breakdown of the different component layers for rendering fields in jcontent. Starting point is the `<Field>` in jcontent https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.jsx
> 
> * `<Field>` (container): takes care of behaviour actions (onChange, onBlur) per field and error handling through formik which calls renderField
> * `renderField()` (in Field): renders input labels, actions, chips badges, etc and determines whether to render
>   `<MultipleField>` and `<SingleField>`
>   
>   * This seems to be the equivalent of the new moonstone Field. Could be good to compare these two, specifically around rendering of errors.
>     Multiple:
> * `<MultipleField>`: coordinates changes between fields (onChange, onFieldRemove, onFieldAdd, onReorder) and renders each field using `OrderableValue`
> * `<OrderableValue>`: handles drag and drop events (using react-dnd) and rendering of each multiple field (wrapped with `<FastField>` (from formik)
>   Single:
> * `<SingleField>`: component that wraps component with `<FastField>` (from formik)

Thanks @gflores-jahia !
I hope our components can handle these use cases, except the orderable value, as we keep the drag-and-drop out of scope for now.
We will put those `Field` components in Beta to allow us to make changes if needed.

PR #468, opened 2024-12-12
MOON-459: Add Spacing pico and update some spacings components

Eevolee:
## JIRA

https://jira.jahia.org/browse/MOON-459

## Description

Add global spacing  `--spacing-pico` & replace `--spacing-nano` with `$spacing-icon` in requiring components style


src/components/ListItem/ListItem.scss
```diff
@@ -33,18 +36,9 @@ $big-image-size: 96px;
 .moonstone-listItem_iconEnd {
     display: flex;
     align-items: center;
-    min-width: 16px;
```
ffffffelix, line 36: Why do you remove the `min-width` ?

src/components/IconTextIcon/IconTextIcon.scss
```diff
@@ -1,11 +1,6 @@
+@use "../../tokens/spacings/spacings" as *;
+
 .moonstone-IconTextIcon {
```
ffffffelix, line 3: I am wondering if we should use we should use a `$spacing-icon_big` spacing/gap when we set `iconSize` to `big`

ffffffelix (commented):
I think we can also use  $spacing-icon_big for some other components like:
- Accordion
- Collapsible

src/components/ListItem/ListItem.scss
Eevolee: Removing it did not cause any incident on related components

src/components/IconTextIcon/IconTextIcon.scss
Eevolee: When icons of `iconTextIcon` are set to size big there is no extra className added to the component, since the spacing is directly assigned as a `gap` on `.moonstone-IconTextIcon` I don't have an efficient way of changing it.

PR #469, opened 2024-12-30
MOON-377: Add border variables for selectors

Eevolee:
## JIRA
https://jira.jahia.org/browse/MOON-377

## Description
Add global border variables for selectors

### DIFF
--- a/src/components/CardSelector/CardSelector.scss
+++ b/src/components/CardSelector/CardSelector.scss
@@ -1,8 +1,10 @@
+@use "../../tokens/borders/borders" as *;
+
 .moonstone-cardSelector {
     width: 100%;
     padding: var(--spacing-small);
 
-    border: 1px solid var(--color-gray40);
+    border: var(--selector-border);
     border-radius: 2px;
     background-color: var(--color-white);
     cursor: pointer;
@@ -10,12 +12,12 @@
     transition: all 0.2s ease-in-out;
 
     &:hover {
-        border-color: var(--color-gray60);
+        border: var(--selector-border_hover);
         box-shadow: 0 1px 6px var(--color-dark20);
     }
 
     &:focus {
-        border-color: var(--color-accent);
+        border: var(--selector-border_focus);
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/components/Dropdown/Dropdown.scss
```diff
@@ -59,16 +58,16 @@ $border-width: 1px;
     }
 
     &.moonstone-opened {
-        box-shadow: inset 0 0 0 $border-width var(--color-accent);
+        border: var(--selector-border);
```
ffffffelix, line 61: You should use `--selector-border_active` instead of `--selector-border` as we want to highlight the dropdown when the component displays its menu.

src/components/Dropdown/Dropdown.scss
```diff
@@ -83,13 +82,13 @@ $border-width: 1px;
...
     color: var(--color-gray_dark);
 
-    box-shadow: inset 0 0 0 $border-width var(--color-gray40);
+    border: var(--selector-border);
 
     &.moonstone-disabled {
         color: var(--color-gray);
 
-        background-color: var(--color-gray_light40);
+        border: var(--selector-border);
```
ffffffelix, line 90: Maybe you don't need to repeat the border here as it is already set with `.moonstone-dropdown_outlined`

src/components/Input/BaseInput/BaseInput.scss
```diff
@@ -82,7 +81,7 @@ $border-radius: 2px;
 
     // Reset Firefox invalid required input style
     &:invalid {
-        box-shadow: none;
```
ffffffelix, line 85: We want to reset the default style provided by Firefox. What is the style added by Firefox when the input is invalid? A `border` or a `box-shadow`?

src/components/CardSelector/CardSelector.scss
```diff
@@ -1,21 +1,23 @@
+@use "../../tokens/borders/borders" as *;
```
ffffffelix, line 1: As you use CSS custom properties, you don't need to import them. They are global.
You only need to import variables with `@use` when you need SCSS variables like `$myvariable`.

Please remove all imports.

ffffffelix (changes requested):

src/components/Input/BaseInput/BaseInput.scss
Eevolee: I did some research and added both `box-shadow: none` and `border: none`

ffffffelix (approved):

PR #470, opened 2024-12-30
MOON-472: Add border radius variables for selectors

Eevolee:
## JIRA
https://jira.jahia.org/browse/MOON-472

## Description
Add globals --selector-border-radius & --selector-border-radius_big

### DIFF
--- a/src/tokens/borders/_borders.scss
+++ b/src/tokens/borders/_borders.scss
@@ -1,6 +1,8 @@
 @forward '../colors/colors';
 
 :root {
+    --selector-border-radius: 2px;
+    --selector-border-radius_big: 4px;
     --selector-border: solid 1px var(--color-gray40);
     --selector-border_hover: solid 1px var(--color-accent);
     --selector-border_focus: solid 1px var(--color-accent);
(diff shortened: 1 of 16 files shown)


src/tokens/borders/_borders.scss
```diff
@@ -1,6 +1,8 @@
 @forward '../colors/colors';
 
 :root {
+    --selector-border-radius: 2px;
+    --selector-border-radius_big: 4px;
```
ffffffelix, line 5: We can go further and always manage all border-radius with variables. I propose:
```
--radius: 4px;
--radius_selector: 2px;
--radius_rounded: 2rem;
```

This naming is not perfect because it mixes the style `rounded` and the usage `selector`, but I hope it will be easier to understand when to use each variables.

ffffffelix (changes requested):

src/tokens/borders/_borders.scss
ffffffelix: I proposed to replace `--selector-border-radius` with `--radius_selector` and `--selector-border-radius_big` with `--radius` and add `--radius_rounded`. Not adding more variables.

In that case, we would have something more generic, and we could use those variables for all components, not only selectors with a shorter naming.

ffffffelix (changes requested):

ffffffelix (approved):

PR #476, opened 2025-01-08
MOON-463: Set background color to parent div of input

Eevolee:
## JIRA

https://jira.jahia.org/browse/MOON-463

## Description

Set `background-color: white` to BaseInput & remove input's default background

### DIFF
--- a/src/components/Input/BaseInput/BaseInput.scss
+++ b/src/components/Input/BaseInput/BaseInput.scss
@@ -12,9 +12,10 @@ $min-width: 40px;
     width: 100%;
     min-width: $min-width;
     min-height: $default-size;
-    background-color: white;
 
     color: var(--color-gray_dark);
+
+    background-color: var(--color-white);
 
     cursor: text;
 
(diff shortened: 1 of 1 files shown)


src/components/Input/BaseInput/BaseInput.scss
```diff
@@ -25,6 +30,12 @@ $min-width: 40px;
         border: var(--border-selector);
         border-radius: var(--radius_small);
     }
+
+    .moonstone-reversed & {
+        // color: var(--color-white);
```
ffffffelix, line 35: Remove the comment please

src/components/Input/BaseInput/BaseInput.scss
```diff
@@ -12,11 +12,16 @@ $min-width: 40px;
     width: 100%;
     min-width: $min-width;
     min-height: $default-size;
+    background-color: white;
```
ffffffelix, line 15: It's better to use color variable even for simple color to be sure to be consitant
```suggestion
    background-color: var(--color-white);
```

ffffffelix (changes requested):

ffffffelix (approved):

PR #478, opened 2025-01-10
MOON-475: Add components Fieldset & DynamicFieldset

Eevolee:
## JIRA
https://jira.jahia.org/browse/MOON-475

## Description
Add components Fieldset & DynamicFieldset

### DIFF
--- a/src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
+++ b/src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
@@ -4,42 +4,41 @@ import '../Fieldset.scss';
 import type {ControlledDynamicFieldsetProps} from './DynamicFieldset.types';
 import {Switch, Typography} from '~/components';
 
-export const ControlledDynamicFieldset = React.forwardRef<HTMLDivElement, ControlledDynamicFieldsetProps>(({
+export const ControlledDynamicFieldset = React.forwardRef<HTMLFieldSetElement, ControlledDynamicFieldsetProps>(({
     id,
     label,
     helper,
     children,
     className,
     buttons,
-    isOpen = false,
+    checked = false,
     onChange,
     ...props
 }, ref) => {
     return (
-        <div
+        <fieldset
             ref={ref}
             id={id}
             className={clsx(
                 'moonstone-dynamic-fieldset',
-                isOpen && 'moonstone-dynamic-fieldset_open',
+                checked && 'moonstone-dynamic-fieldset_open',
                 'flexCol_nowrap',
                 className
             )}
+            aria-checked={checked}
             {...props}
         >
-            <div className={clsx('flexRow_nowrap', 'flexFluid', 'alignCenter')}>
-                <div className="flexRow_nowrap flexFluid alignCenter">
-                    <Typography isNowrap component="label" variant="heading" weight="bold">{label}</Typography>
-                </div>
-                <Switch checked={isOpen} onChange={onChange}/>
+            <legend className={clsx('flexRow_nowrap', 'flexFluid', 'alignCenter')}>
+                <Typography isNowrap className="flexRow_nowrap flexFluid alignCenter" component="label" variant="heading" weight="bold">{label}</Typography>
+                <Switch id="moonstone-dynamic-fieldset-switch" checked={checked} onChange={onChange}/>
                 {buttons && buttons}
-            </div>
+            </legend>
             {helper &&
                 <Typography variant="caption" className={clsx('moonstone-fieldset_helper')}>{helper}</Typography>}
             <div className={clsx('moonstone-fieldset_children', 'flexCol_nowrap')}>
-                {isOpen && children}
+                {(checked && children) && children}
             </div>
-        </div>
+        </fieldset>
     );
 });
 
--- a/src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
+++ b/src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
@@ -35,9 +35,10 @@ export const ControlledDynamicFieldset = React.forwardRef<HTMLFieldSetElement, C
             </legend>
             {helper &&
                 <Typography variant="caption" className={clsx('moonstone-fieldset_helper')}>{helper}</Typography>}
+            {(checked && children) &&
             <div className={clsx('moonstone-fieldset_children', 'flexCol_nowrap')}>
-                {(checked && children) && children}
-            </div>
+                {children}
+            </div>}
         </fieldset>
     );
 });
--- a/src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
+++ b/src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import clsx from 'clsx';
+import '../Fieldset.scss';
+import type {ControlledDynamicFieldsetProps} from './DynamicFieldset.types';
+import {Switch, Typography} from '~/components';
+
+export const ControlledDynamicFieldset = React.forwardRef<HTMLDivElement, ControlledDynamicFieldsetProps>(({
+    id,
+    label,
+    helper,
+    children,
+    className,
+    buttons,
+    isOpen = false,
+    onChange,
+    ...props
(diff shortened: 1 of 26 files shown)


src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
```diff
@@ -0,0 +1,46 @@
...
+            className={clsx(
+                'moonstone-dynamic-fieldset',
+                isOpen && 'moonstone-dynamic-fieldset_open',
+                'flexCol_nowrap',
+                className
+            )}
+            {...props}
+        >
+            <div className={clsx('flexRow_nowrap', 'flexFluid', 'alignCenter')}>
+                <div className="flexRow_nowrap flexFluid alignCenter">
```
ffffffelix, line 31: Not sure why this `div` is mandatory

src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    <Typography isNowrap component="label" variant="heading" weight="bold">{label}</Typography>
```
ffffffelix, line 32: You can use `legend` tag instead to set the title of a `fieldset` ([MDN doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend))

src/components/Fieldset/DynamicFieldset/DynamicFieldset.types.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Dynamic fieldset helper
+     */
+    helper?: string;
+
+    /**
+     * Define fieldset field(s)
+     */
+    children: React.ReactElement;
```
ffffffelix, line 27: I checked, and there are some cases when the fieldset component is used as a boolean, so no children are provided.
So `children` become an optional prop.

src/components/Fieldset/DynamicFieldset/DynamicFieldset.types.ts
```diff
@@ -0,0 +1,61 @@
...
+     * Dynamic fieldset action(s)
+    */
+   buttons?: React.ReactElement;
+};
+
+type ControlledProps = {
+    /**
+     * Whether dynamic fieldset is open or not
+     */
+    isOpen: boolean;
```
ffffffelix, line 39: I prefer `checked` to match the `Switch` component props.

src/components/Fieldset/DynamicFieldset/DynamicFieldset.types.ts
```diff
@@ -0,0 +1,61 @@
...
+     * Dynamic fieldset's function onChange
+     */
+    onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+type UncontrolledProps = {
+    /**
+     * Whether dynamic fieldset is open by default
+     */
+    defaultOpen?: boolean;
```
ffffffelix, line 51: So `defaultChecked` to matches the `Switch` component

src/components/Fieldset/DynamicFieldset/DynamicFieldset.types.ts
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
+
+type BaseProps = {
```
ffffffelix, line 3: To avoid repeating all props here, you can import and inherit from `FieldsetProps` as both shared the same based props

src/components/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,42 @@
...
+    id,
+    label,
+    helper,
+    children,
+    className,
+    buttons,
+    ...props
+}, ref) => {
+    return (
+        <div
```
ffffffelix, line 17: Use a `fieldset` tag in order to have HTML more semantic

src/components/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,42 @@
...
+                <div className="flexRow_nowrap flexFluid alignCenter">
+                    <Typography isNowrap component="label" variant="heading" weight="bold">{label}</Typography>
+                </div>
+                {buttons && buttons}
+            </div>
+            {helper &&
+                <Typography variant="caption" className={clsx('moonstone-fieldset_helper')}>{helper}</Typography>}
+            <div className={clsx('moonstone-fieldset_children', 'flexCol_nowrap')}>
+                {children}
+            </div>
```
ffffffelix, line 37: As `children` becomes optional, you must not display this `div` when `children` is not provided

src/components/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,42 @@
...
+            id={id}
+            className={clsx(
+                'moonstone-fieldset',
+                'flexCol_nowrap',
+                className
+            )}
+            {...props}
+        >
+            <div className={clsx('flexRow_nowrap', 'flexFluid', 'alignCenter')}>
+                <div className="flexRow_nowrap flexFluid alignCenter">
```
ffffffelix, line 28: I think you can remove this `div`, you can add classes on `Typography` is it's needed

src/components/Fieldset/Fieldset.types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Fieldset helper
+     */
+    helper?: string;
+
+    /**
+     * Define fieldset field(s)
+     */
+    children: React.ReactElement;
```
ffffffelix, line 27: `children` becomes optional

src/components/Fieldset/DynamicFieldset/DynamicFieldset.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {StoryObj, Meta} from '@storybook/react';
+
+import {DynamicFieldset} from './index';
+import {Field, FieldSelector} from '~/components';
+import markdownNotes from './DynamicFieldset.md';
+import {Button, Chip, Input} from '~/components';
+import {Add, Language, MoreVert} from '~/icons';
+
+const meta: Meta<typeof DynamicFieldset> = {
+    title: 'Components/DynamicFieldset',
```
ffffffelix, line 11: I think it will be better to put `DynamicFieldset` in the same "folder" than the `Fieldset` component in storybook to keep both closer

src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
```diff
@@ -0,0 +1,46 @@
...
+    className,
+    buttons,
+    isOpen = false,
+    onChange,
+    ...props
+}, ref) => {
+    return (
+        <div
+            ref={ref}
+            id={id}
```
ffffffelix, line 21: You also have to add some `data-` or `aria-` attributes to match the state in order to be more semantic and help our tests with attributes to target

ffffffelix (changes requested):
It will be great to clicking on the label or description to change the switch.
You can do it with a `label` with a `for` attribute which match the `id` of the Switch.

src/components/Fieldset/DynamicFieldset/ControlledDynamicFieldset.tsx
```diff
@@ -4,42 +4,41 @@ import '../Fieldset.scss';
...
-                <Switch checked={isOpen} onChange={onChange}/>
+            <legend className={clsx('flexRow_nowrap', 'flexFluid', 'alignCenter')}>
+                <Typography isNowrap className="flexRow_nowrap flexFluid alignCenter" component="label" variant="heading" weight="bold">{label}</Typography>
+                <Switch id="moonstone-dynamic-fieldset-switch" checked={checked} onChange={onChange}/>
                 {buttons && buttons}
-            </div>
+            </legend>
             {helper &&
                 <Typography variant="caption" className={clsx('moonstone-fieldset_helper')}>{helper}</Typography>}
             <div className={clsx('moonstone-fieldset_children', 'flexCol_nowrap')}>
```
ffffffelix, line 38: The condition `checked && children` should also wrap the `div`, to avoid having an empty `div` when no `children` is provided or when when the fieldset is collapsed.

ffffffelix (changes requested):

ffffffelix (approved):

PR #492, opened 2025-01-29
MOON-483: Improve tests coverage

Eevolee:
## JIRA
https://jira.jahia.org/browse/MOON-483

## Description
Improve components' tests coverage by adding new tests, semantic attributes and lightly fixing some components

### DIFF
--- a/src/components/Table/SortIndicator/SortIndicator.tsx
+++ b/src/components/Table/SortIndicator/SortIndicator.tsx
@@ -18,11 +18,11 @@ export const SortIndicator: React.FC<SortIndicatorProps> = ({
     );
 
     if (direction === 'descending') {
-        return <ArrowDown {...props} className={classNameProps}/>;
+        return <ArrowDown {...props} aria-label="arrowDown-icon" className={classNameProps}/>;
     }
 
     if (direction === 'ascending') {
-        return <ArrowUp {...props} className={classNameProps}/>;
+        return <ArrowUp {...props} aria-label="arrowUp-icon" className={classNameProps}/>;
     }
 };
 
--- a/src/components/Table/TablePagination/TablePagination.tsx
+++ b/src/components/Table/TablePagination/TablePagination.tsx
@@ -60,6 +60,7 @@ export const TablePagination: React.FC<TablePaginationProps> = ({
                 className={clsx(cssClass32)}
                 icon={<ChevronFirstPage/>}
                 variant="ghost"
+                aria-label="go-to-first-page"
                 data-sel-role="table-pagination-button-first-page"
                 isDisabled={currentPage === 1}
                 onClick={() => onPageChange(1)}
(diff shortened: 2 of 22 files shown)


src/components/Table/TablePagination/TablePagination.tsx
```diff
@@ -60,6 +60,7 @@ export const TablePagination: React.FC<TablePaginationProps> = ({
                 className={clsx(cssClass32)}
                 icon={<ChevronFirstPage/>}
                 variant="ghost"
+                aria-label="go-to-first-page"
```
ffffffelix, line 63: `aria-label` is mainly for assistive technology, so you should write `Go to the first page`. But this string will not be translated so maybe it will be better to use `data-sel-role` to target element in our tests

src/components/Table/SortIndicator/SortIndicator.tsx
```diff
@@ -18,11 +18,11 @@ export const SortIndicator: React.FC<SortIndicatorProps> = ({
     );
 
     if (direction === 'descending') {
-        return <ArrowDown {...props} className={classNameProps}/>;
+        return <ArrowDown {...props} aria-label="arrowDown-icon" className={classNameProps}/>;
```
ffffffelix, line 21: As `aria-label` is for accessible name, you can be more descriptive: `Icon for sorting in descending order`

src/components/Table/SortIndicator/SortIndicator.tsx
```diff
@@ -18,11 +18,11 @@ export const SortIndicator: React.FC<SortIndicatorProps> = ({
...
     }
 
     if (direction === 'ascending') {
-        return <ArrowUp {...props} className={classNameProps}/>;
+        return <ArrowUp {...props} aria-label="arrowUp-icon" className={classNameProps}/>;
```
ffffffelix, line 25: As `aria-label` is for accessible name, you can be more descriptive: `Icon for sorting in ascending order`

ffffffelix (changes requested):

src/components/PrimaryNav/PrimaryNavItem/PrimaryNavItem.spec.tsx
```diff
@@ -3,6 +3,7 @@ import {render, screen} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {PrimaryNavItem} from './index';
 import {PrimaryNav} from '../PrimaryNav';
+import {Badge} from '~/main';
```
ffffffelix, line 6: As `main.ts` is only there for compatibility reasons, I prefer to use

```suggestion
import {Badge} from '~/components';
```

ffffffelix (changes requested):

ffffffelix (approved):

PR #493, opened 2025-01-29
QA-15352: Fix dnd issue for ListSelector

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/QA-15352

## Description

- Add missing parameter on `onDragOver` event
- Check for falsey values in `value` instead of just undefined (it can also come back as null and throw errors)


ffffffelix:
Is it possible to add tests to check this behavior?

gflores-jahia:
Unfortunately, I don't see a way to test the event handlers with jest. Maybe with integration test like cypress, but we've also not had any luck with that on jcontent side.

ffffffelix (approved):

PR #973, opened 2025-03-03
chore: unify ghost variant styling

Eevolee:
## Description

Add global placeholder `%variant_ghost` to unify the styling of ghost components & remove unused props color & variant on TabItem


src/globals/_variants.scss
```diff
@@ -0,0 +1,146 @@
...
+
+        border-color: transparent;
+        background-color: transparent;
+        cursor: not-allowed;
+        opacity: 0.4;
+    }
+}
+
+%variant_ghost_accent {
+    @extend %variant_ghost;
```
ffffffelix, line 61: As `%variant_ghost` manages state `:hover`, `:focus` , `:active`. I don't think we need to extend it. Plus, it will almost be overridden.

src/globals/_variants.scss
```diff
@@ -0,0 +1,146 @@
...
+            color: var(--color-white);
+
+            border-color: transparent;
+            background-color: var(--color-accent_dark);
+        }
+    }
+}
+
+%variant_ghost_danger {
+    @extend %variant_ghost;
```
ffffffelix, line 105: Same, do we need to extend it, as we redefine all styles ?

src/components/Dropdown/Dropdown.scss
```diff
@@ -61,14 +63,14 @@
...
+    // }
 
-    &:focus {
-        border: var(--border-selector_focus);
-        outline: none;
-    }
+    // &:focus {
+    //     border: var(--border-selector_focus);
+    //     outline: none;
+    // }
```
ffffffelix, line 73: If you don't need ti anymore, can you remove it please ?

src/components/Button/Button.scss
```diff
@@ -1,58 +1,45 @@
 @use "variants/default";
-@use "variants/ghost";
 @use "variants/outlined";
 @use "../../tokens/spacings/spacings" as *;
+@use "../../globals/variants";
 
+$colors: "default", "accent", "danger";
```
ffffffelix, line 6: What is the purpose of the variable `$colors`? It seems you don't use it.

ffffffelix (changes requested):

ffffffelix (approved):

PR #979, opened 2025-03-06
feat(component): add Tooltip with new dependency react floating UI

Eevolee:
## Description

Add new dependency `@floating-ui/react`
Add new component `Tooltip`

### DIFF
--- a/src/components/Tooltip/Tooltip.scss
+++ b/src/components/Tooltip/Tooltip.scss
@@ -1,0 +1,21 @@
+$bgColor: var(--color-gray_dark);
+
+.moonstone-tooltip_label {
+    z-index: 1;
+
+    width: inherit;
+    padding: var(--spacing-nano) var(--spacing-small);
+
+    font-size: 90%;
+    color: var(--color-white);
+    text-overflow: clip;
+
+    border-radius: var(--radius);
+    background-color: $bgColor;
+}
+
+.moonstone-tooltip_arrow {
+    z-index: 1;
+
+    fill: $bgColor;
+}
--- a/src/components/Tooltip/Tooltip.scss
+++ b/src/components/Tooltip/Tooltip.scss
@@ -4,11 +4,10 @@ $bgColor: var(--color-gray_dark);
     z-index: 1;
 
     width: inherit;
+    max-width: 450px;
     padding: var(--spacing-nano) var(--spacing-small);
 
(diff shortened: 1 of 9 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,70 @@
...
+                className={clsx('moonstone-tooltip', className)}
+                {...getReferenceProps()}
+                {...props}
+            >
+                {/* Children is the element to interact with in order for the tooltip to appear */}
+                {React.cloneElement(children, {
+                'aria-describedby': 'moonstone-tooltip_label'
+                })}
+                {isOpen &&
+                <span
```
ffffffelix, line 54: Can you use the `Typography` component instead of the `span` ?
By utilizing the typography component, we will ensure a coherent style for the text.

src/components/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,21 @@
...
+
+.moonstone-tooltip_label {
+    z-index: 1;
+
+    width: inherit;
+    padding: var(--spacing-nano) var(--spacing-small);
+
+    font-size: 90%;
+    color: var(--color-white);
+    text-overflow: clip;
```
ffffffelix, line 11: If we use the `Typography` component we won't need to set `font-size` and `text-overflow`

src/components/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,21 @@
+$bgColor: var(--color-gray_dark);
+
+.moonstone-tooltip_label {
+    z-index: 1;
+
+    width: inherit;
```
ffffffelix, line 6: Maybe we should add a `max-width` to prevent tooltip with long text to take the full width of the window
I suggest something like `max-width: 450px`

src/components/Tooltip/Tooltip.tsx
Eevolee: It creates an error because the tooltip's position depends on `ref={refs.setFloating}` but ref does not exist in Typography's props

Eevolee: I can replace the span with a div and wrap `{label}` in `Typography` though

ffffffelix (approved):

PR #1012, opened 2025-04-10
fix: typing issues and React import

ffffffelix:
### React Import
- Remove unnecessary React import
- Update SVGR configuration
- Add `plugin:react/jsx-runtime` to our eslint configuration to prevent lint error about missing React import

### Typing
- Update tests to use correct typing when we use our component
- `isReversed` becomes optional in `AccordionItem` as all other components
- `thumbnailType` becomes optional in `CardSelector` as we set a default value
- `id` becomes optional when it is not used to fill other attributes
- Refactor typing of `DropdownData` to fit all use cases
- Refactor `DropdownMenuProps`, `TreeViewMenuProps` and `MenuProps` to mutualize types
- Update `children` type to allow one or many items

## Tests

- [X] Unit Tests
- [ ] Accessibility is OK


src/components/Badge/Badge.tsx
```diff
@@ -1,10 +1,10 @@
-import React, {FunctionComponent} from 'react';
+import React from 'react';
```
GauBen, line 1:
```suggestion
import type React from 'react';
```

You can also rewrite as `import type { FC } from 'react';`

src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -102,7 +101,7 @@ describe('Checkbox', () => {
     it('should have mixed state when specified with the isIndeterminate prop', () => {
         render(
             <Checkbox
-                checked="mixed"
+                indeterminate
```
GauBen, line 104: Cool finding!

src/components/Dropdown/Dropdown.types.ts
```diff
@@ -17,22 +17,28 @@ export type DropdownDataOption = {
...
-    groupLabel?: string;
-    options?: [DropdownDataOption];
 }
 
+export type DropdownDataGrouped = {
+    groupLabel: string;
+    options: DropdownDataOption[];
+}
+
+export type DropdownData = DropdownDataOption[] | DropdownDataGrouped[];
```
GauBen, line 27: If the two things can be interleaved, it should be `Array<DropdownDataOption | DropdownDataGrouped>`

Otherwise, it's good as it is right now

src/components/Fieldset/Fieldset.types.ts
```diff
@@ -24,7 +24,7 @@ export type FieldsetProps = {
     /**
      * Define fieldset field(s)
      */
-    children: React.ReactElement;
+    children: React.ReactElement | React.ReactElement[];
```
GauBen, line 27: You may write `React.ReactNode` here to accept all JSX nodes

src/components/Dropdown/Dropdown.types.ts
ffffffelix: It's exclusive `DropdownData` can be `DropdownDataOption[]` OR `DropdownDataGrouped[]`

GauBen (approved):

PR #1024, opened 2025-04-28
fix: Storybook accessibility fix

Eevolee:
Fix most of accessibility issues found in storybook except colors contrast issues


src/components/Menu/Menu.tsx
```diff
@@ -112,12 +112,13 @@ export const Menu: React.FC<MenuProps> = ({
...
                 ref={itemRef}
                 style={styleMenu}
-                role="listbox"
+                role="list"
                 className={clsx(
                     'moonstone-menu',
                     className,
                     {'moonstone-hidden': !isDisplayed || !stylePosition}
                 )}
+                aria-label="moonstone-menu"
```
ffffffelix, line 121: Same as other

src/components/SecondaryNav/SecondaryNavHeader/SecondaryNavHeader.tsx
```diff
@@ -5,7 +5,8 @@ import {Typography} from '~/components/Typography';
 import {SecondaryNavHeaderProps} from './SecondaryNavHeader.types';
 
 export const SecondaryNavHeader: React.FC<SecondaryNavHeaderProps> = ({children}) => (
-    <Typography component="header"
+    <Typography component="div"
+                aria-label="moonstone-secondaryNavHeader"
```
ffffffelix, line 9: Are you sure we can't use a `header` tag here?
Do we need a `aria-label` for a `div` ?

Eevolee: `component="header"` and `role="banner"` both result in accessibility error "Banner landmark should not be contained in another landmark".
The aria-label is required for a test in `SecondaryNav.spec.tsx`

ffffffelix (approved):

PR #1033, opened 2025-05-07
feat: CardSelector improvements

ffffffelix:
## Description

- Increase the `height` of the `CardSelector`
- Do not manage `cursor` anymore. It will be the responsibility of the consumer of the library as it depends on the usage of the components
- Truncate too long `displayName` and `systemName`
- Review the shadow and border color

### DIFF
--- a/src/components/CardSelector/CardSelector.tsx
+++ b/src/components/CardSelector/CardSelector.tsx
@@ -90,7 +90,7 @@ export const CardSelector = React.forwardRef<HTMLDivElement, CardSelectorProps>(
                     {displayName && (
                         <Typography
                             isNowrap
-                            id={`${id}-displayName`}
+                            id={id && `${id}-displayName`}
                             className={clsx('moonstone-cardSelector_displayName')}
                             variant="body"
                             component="span"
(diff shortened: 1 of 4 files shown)


src/components/CardSelector/CardSelector.tsx
```diff
@@ -114,16 +116,15 @@ export const CardSelector = React.forwardRef<HTMLDivElement, CardSelectorProps>(
                 </div>
                 {(chips || information) && (
                     <div className={clsx('flexRow_nowrap')}>
-                        {chips && (
-                            chips.map(chip => (chip))
-                        )}
+                        {chips?.map(chip => (chip))}
```
gflores-jahia, line 119: I believe we can  just render this directly if it's already a list of react components (assuming there's already a `key` param added to the components), no need for the map:

```suggestion
                        {chips}
```

gflores-jahia (approved):

src/components/CardSelector/CardSelector.tsx
ffffffelix: Right, simpler is better !

gflores-jahia (approved):

PR #1047, opened 2025-05-16
feat: add onArrowNavigation to manage navigation with ArrowDown and ArrowUp

Eevolee:
- Add `onArrowNavigation` hook
- Add `onArrowNavigation` tests
- Implement `onArrowNavigation` list/tree/menu item type elements


Eevolee:
- Began implementing on `ControlledTreeView` but the result is very random
- Needs to be implemented on `ListItem` but conflict between the component's `HTMLLIElement` ref & the hook's `Element` ref prop

ffffffelix (approved):

PR #1051, opened 2025-05-22
feat: allow icon component as thumbnail

ffffffelix:
## Description

- Rename `thumbnailURL` to `thumbnail`
- Allow passing an icon component as a thumbnail
- UPdate stories
- Update the test

## Tests

- [X] Unit Tests
- [ ] Accessibility is OK

### DIFF
--- a/src/components/CardSelector/CardSelector.tsx
+++ b/src/components/CardSelector/CardSelector.tsx
@@ -2,16 +2,42 @@ import React from 'react';
 import clsx from 'clsx';
 
 import './CardSelector.scss';
-import type {CardSelectorProps} from './CardSelector.types';
+import type {CardSelectorProps, ThumbnailProps} from './CardSelector.types';
 import {Typography} from '~/components';
 import {FileBroken, Image} from '~/icons/components';
+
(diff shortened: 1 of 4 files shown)


src/components/CardSelector/CardSelector.tsx
```diff
@@ -2,16 +2,42 @@ import React from 'react';
 import clsx from 'clsx';
 
 import './CardSelector.scss';
-import type {CardSelectorProps} from './CardSelector.types';
+import type {CardSelectorProps, ThumbnailProps} from './CardSelector.types';
 import {Typography} from '~/components';
 import {FileBroken, Image} from '~/icons/components';
 
+const ThumbnailCmp: React.FC<ThumbnailProps> = ({thumbnail, thumbnailType, thumbnailAlt}) => {
+    console.log(thumbnail);
```
gflores-jahia, line 10: might be better use debug instead of log if we want to keep, or remove

ffffffelix: oups I forgot to remove it

PR #1072, opened 2025-06-17
chore: update Tabs to meet W3C standards

Eevolee:
- Add horizontal arrow navigation using `onArrowNavigation` hook
- Add role 'tablist' on `Tab` & 'tab' on `TabItem`
- Add attribute `aria-selected` on `TabItem`

### DIFF
--- a/src/components/Tab/TabItem/TabItem.tsx
+++ b/src/components/Tab/TabItem/TabItem.tsx
@@ -37,8 +37,8 @@ export const TabItem: React.FC<TabItemProps> = ({
                 'aria-selected': isSelected,
                 disabled: isDisabled,
                 onClick,
-                ...props,
-                ...onArrowNavigation({ref: ref, direction: 'horizontal'})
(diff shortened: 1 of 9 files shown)


src/components/Tab/TabItem/TabItem.tsx
```diff
@@ -15,29 +16,35 @@ export const TabItem: React.FC<TabItemProps> = ({
...
+                    'alignCenter',
+                    className
+                ),
+                ref: ref,
+                role: 'tab',
+                'aria-selected': isSelected,
+                disabled: isDisabled,
+                onClick,
+                ...props,
+                ...onArrowNavigation({ref: ref, direction: 'horizontal'})
```
ffffffelix, line 41: I think it's better to add `..props` at the end to easily override attributes returned by `onArrowNavigation`, otherwise people who use that component cannot set `tabIndex`.

ffffffelix (changes requested):

ffffffelix (approved):

PR #1084, opened 2025-06-26
feat: add is-focused mixin to unify focusable components style

Eevolee:
- Add `%is-focused` mixin to `/utils/_mixins.scss`
- Add `%is-focused` to unify focus style on focusable components

### DIFF
--- a/src/components/Accordion/AccordionItem/AccordionItem.scss
+++ b/src/components/Accordion/AccordionItem/AccordionItem.scss
@@ -1,4 +1,5 @@
 @use "../../../tokens/spacings/spacings" as *;
+@use "../../../utils/mixins" as *;
 
 $accordionHeader_height: 48px;
 
@@ -21,6 +22,10 @@ $accordionHeader_height: 48px;
     &.moonstone-reversed {
         border-color: var(--color-gray);
     }
+
+    &:has(.moonstone-accordionItem_header:focus-visible) {
+        @extend %is-focused;
+    }
 }
 
 // ---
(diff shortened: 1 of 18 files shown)


src/components/Accordion/AccordionItem/AccordionItem.scss
```diff
@@ -21,6 +22,10 @@ $accordionHeader_height: 48px;
     &.moonstone-reversed {
         border-color: var(--color-gray);
     }
+
+    &:has(.moonstone-accordionItem_header:focus-visible) {
```
ffffffelix, line 26: Should it be better to apply the style directly to `moonstone-accordionItem_header` instead of the parent?
In that case we will have to override `outline-offset` to `-2px`

src/utils/_mixins.scss
```diff
@@ -19,3 +19,9 @@
     cursor: not-allowed;
     opacity: 0.6;
 }
+
+%is-focused {
+    border-radius: 4px;
```
ffffffelix, line 24: I am wondering if it would be better to remove `border-radius` and use the radius of the item instead. It will prevent overriding this style in some cases (eg: PrimaryNavItem)

src/components/Button/variants/_outlined.scss
```diff
@@ -48,9 +48,7 @@
 
         &:hover:not(:disabled),
         &:focus-visible:not(:disabled) {
-            color: var(--color-accent_dark);
-
-            border-color: var(--color-accent_dark);
+            @extend %is-focused;
```
ffffffelix, line 51: Be careful, you also apply the focus style when the item is hovering

src/components/Button/variants/_outlined.scss
```diff
@@ -72,9 +70,7 @@
 
         &:hover:not(:disabled),
         &:focus-visible:not(:disabled) {
-            color: var(--color-danger);
-
-            border-color: var(--color-danger);
+            @extend %is-focused;
```
ffffffelix, line 73: same here

ffffffelix (changes requested):
I guess you missed to update https://github.com/Jahia/moonstone/blob/develop/src/globals/_variants.scss

PR #1097, opened 2025-07-17
feat: fixes and improvements for CE integration

r3dm1ke:
## Description

With this PR (https://github.com/Jahia/jcontent/pull/1893/files) I updated the disabled styles in jcontent, as well as switched from using MUI inputs in CE to fully using Moonstone inputs for everything. 

- Postfix components needed for inputs with pickers (color picker/date picker)
- The focus styling fix for textareas and outlined inputs is needed because once input is focused, it displays a large thick blue border around the input, margined as well, so it is even bigger than the input. You could extend this fix to all variants of input if needed, but jcontent is only using outlined
- 

## Tests

- [ ] Unit Tests
- [ ] Accessibility is OK

## Checklist

- [ ] Example in storybook
- [ ] Allows custom props

### DIFF
--- a/src/components/Input/BaseInput/BaseInput.scss
+++ b/src/components/Input/BaseInput/BaseInput.scss
@@ -24,17 +24,21 @@ $min-width: 40px;
         @extend %selector_disabled;
     }
 
-    &:has(input:focus-visible) {
-        @extend %is-focused;
-    }
-
     &.moonstone-big {
         min-height: $big-size;
+    }
+
+    &:has(input:focus-visible) {
+        @extend %is-focused;
     }
 
     &.moonstone-outlined {
         border: var(--border-selector);
         border-radius: var(--radius-selector);
+
+        &:hover {
+            border: var(--border-selector_hover);
+        }
     }
 
     .moonstone-reversed & {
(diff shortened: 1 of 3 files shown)


gflores-jahia (approved):
Could use some description on the changes.

To me it looks ok, but let's wait for Felix feedback about adding postfixComponents

ffffffelix:
Could you give me more context about these changes:
- Why do we need `postfixComponents `?
- Some changes had already been made regarding the disabled state with https://github.com/Jahia/moonstone/pull/1091

src/components/Input/BaseInput/BaseInput.scss
```diff
@@ -24,22 +24,30 @@ $min-width: 40px;
...
-
     &.moonstone-big {
         min-height: $big-size;
     }
 
     &.moonstone-outlined {
         border: var(--border-selector);
         border-radius: var(--radius-selector);
+
+        &:hover {
```
ffffffelix, line 35: Nice catch! It's better with `hover` style 👍

src/components/Input/BaseInput/BaseInput.types.ts
```diff
@@ -62,6 +62,11 @@ type BasicBaseInputProps = Omit<React.ComponentPropsWithRef<'input'>, 'size' | '
      */
     prefixComponents?: React.ReactElement[];
 
+    /**
+     * Component
```
ffffffelix, line 66: Can you change the description to be more descriptive? And we also should rephrase the description of `prefixComponents`

ffffffelix (changes requested):

ffffffelix (approved):

PR #1124, opened 2025-08-07
chore: add Dropdown prop isLoading

Eevolee:
- Add prop `isLoading` to `Dropdown`

### DIFF
--- a/src/components/Dropdown/Dropdown.scss
+++ b/src/components/Dropdown/Dropdown.scss
@@ -21,7 +21,8 @@
     gap: var(--spacing-nano) var(--spacing-small);
 }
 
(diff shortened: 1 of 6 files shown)


ffffffelix (changes requested):
few feedbacks:
- Add `aria-busy={isLoading ? 'true' : undefined}` on the dropdown when the component is loading this attribute must not be present when the component is not loading.
- Prevent `:hover`  and click when the component is loading
- Add `isLoading` to the story `TemplateSimple` in order to play with the `isLoading` prop in storybook UI
- Add a test about `isLoading`

ffffffelix (approved):

PR #1126, opened 2025-08-11
fix: remove focus style on disabled, readonly & loading components

Eevolee:
- Remove focus style on disabled, readonly & loading components


ffffffelix (changes requested):
- We miss to prevent focus for `EmptyCardSelector` when it `disabled` and/or `readonly`
- We might manage `focusOnField` to prevent to focus when the Input is `disabled` or `readonly`

ffffffelix (approved):

PR #1137, opened 2025-09-18
feat: Add NumberInput component

### LINKED ISSUE #1129: Add number capability to Input component
### Story
As a user
I want to to be able to user number input with the possibility of having decimals and no alphabetic chars

### Test Strategy
Cypress tests?

### Tech Notes

@jahia/design-system allowed this type of behaviour by using react-number-format, I had to patch jcontent to provide similar functionality as material ui components are no longer desirable. It would be great to have this functionality come directly from Moonstone.

Also, note that updating to 5.x of `react-number-format` breaks jcontent, even if that input is never used.

### Labels 
En: 
Fr: 
De:

abourgoindev:
- Add `NumberInput` component, stories & tests
- Add `filterInput` util function & tests
- Add `onArrowIncrementation` hook & tests
- Modifications on `BaseInput` component in order to be implemented into `NumberInput`


src/components/Input/BaseInput/ControlledBaseInput.tsx
```diff
@@ -28,22 +28,32 @@ export const ControlledBaseInput: React.FC<ControlledBaseInputProps> = ({
...
     const classNameProps = clsx('moonstone-baseInput', `moonstone-${size}`, `moonstone-${variant}`, className);
+    let handleClear = onClear || null;
 
     useEffect(() => {
         if (focusOnField && !isDisabled && !isReadOnly) {
             inputRef.current.focus({preventScroll: true});
         }
     }, [focusOnField, isDisabled, isReadOnly]);
 
+    // Merging refs because SearchInput needs inputRef but NumberInput needs ref
```
ffffffelix, line 43: As a future improvement, do you think we can update `SearchInput` to use the same `ref` to remove this part?

src/components/Input/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+    // Filters the value on mount
+    useEffect(() => {
+        const initialValue = inputRef.current?.value;
+
+        if (inputRef.current) {
+            const filteredValue = filterInput(initialValue, allowNegative, allowDecimal);
+
+            if (filteredValue !== initialValue) {
+                inputRef.current.value = filteredValue;
+                inputRef.current.defaultValue = filteredValue;
```
ffffffelix, line 41: Do we always need to set `value` and `defaultValue` ?

src/utils/inputFilter.ts
```diff
@@ -0,0 +1,17 @@
+export const filterInput = (value: string, allowNegative: boolean, allowDecimal = false) => {
```
ffffffelix, line 1: To maintain consistency, the function name and the file should have the same name. I propose `filterInputValue` to highlight the fact that we want to filter the value.

src/components/Input/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,76 @@
+import React, {useRef, useEffect} from 'react';
+import type {NumberInputProps} from './NumberInput.types';
+import {BaseInput} from '../BaseInput';
+import {onArrowIncrementation} from '~/hooks';
+import {filterInput} from '~/utils/inputFilter';
+
+const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(({
+    allowNegative = true,
```
ffffffelix, line 8: Finally, I changed my mind; I think it's better to set `allowNegative` to `false` by default, as I believe it is the most common usage of a number input.

ffffffelix (changes requested):

ffffffelix (changes requested):

gflores-jahia (commented):
looks ok to me.

The only thing I didn't see (or maybe I missed) is the ability to specify decimal separator like [here](https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/SelectorTypes/Text/Text.jsx#L40) which changes depending on the uilang being used

src/components/Input/BaseInput/UncontrolledBaseInput.tsx
```diff
@@ -1,19 +1,26 @@
...
+export const UncontrolledBaseInput = React.forwardRef<HTMLInputElement, UncontrolledBaseInputProps>(({defaultValue, onChange, hasFilteredValue = false, allowNegative, allowDecimal, separator, ...props}, ref) => {
     const [inputValue, setBaseInputValue] = useState(defaultValue);
 
     const handleOnChange = (event: ChangeEvent<HTMLInputElement>) => {
-        setBaseInputValue(event.target.value);
+        let value = event.target.value;
+
+        if (hasFilteredValue) {
+            value = filterInputValue(event.target.value, allowNegative, allowDecimal, separator);
+        }
```
ffffffelix, line 14: Instead of using `hasFilteredValue` to identify if you need to filter the value, hardcode the filtering into the function. I suggest replacing this prop with `filteredFunction` to pass the function to the component. 
In that case, we will be more future-proof to pass a custom filtered function.

In that case, don't forget to update omit types.

src/components/Input/BaseInput/BaseInput.types.ts
```diff
@@ -115,6 +115,26 @@ type UncontrolledProps = {
...
+
+    /**
+     * Allows input value to be decimal
+     */
+    allowDecimal?: boolean;
+
+    /**
+     * Character to use as a separator
+     */
+    separator?: string;
```
ffffffelix, line 137: Restrict to `.` and `,` instead of authorizing all strings.

ffffffelix (changes requested):

ffffffelix (approved):

PR #1143, opened 2025-09-29
chore: Migrate dropdown story in typescript

BouklifYacine:
Migrated dropdown stories from jsx to tsx


src/components/Dropdown/Dropdown.types.ts
```diff
@@ -3,10 +3,16 @@ import type {TreeViewData} from '~/components/TreeView/TreeView.types';
 
 type DropdownVariant = 'ghost' | 'outlined';
 
-type DropdownSize = 'small' | 'medium';
+export type DropdownSize = 'small' | 'medium';
```
ffffffelix, line 6: If you use' size="medium" ' instead of `size={'medium' as DropdownSize}`, you shouldn't need to export this type.

src/components/Dropdown/Dropdown.types.ts
```diff
@@ -24,12 +30,33 @@ export type DropdownDataGrouped = {
     options: DropdownDataOption[];
 }
 
+export type DropdownPill = DataLanguageType[]
```
ffffffelix, line 33: You don't need to create custom type for a specific story. All Dropdown data must use `DropdownData` (`DropdownDaraOption[]` or `DropdownDataGrouped[]`) or `DropdownDataTree`

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -383,8 +363,8 @@ export const WithDescription = {
         return (
             <Dropdown
                 {...args}
-                label={currentOption.label}
-                value={currentOption.value}
+                label={currentOption?.label || 'Select something'}
```
ffffffelix, line 366: - You can remove label
- You can use `placeholder` to set `Select Something`
- Use `value={currentOption?.value}`

src/components/Dropdown/Dropdown.types.ts
```diff
@@ -8,7 +8,7 @@ type DropdownSize = 'small' | 'medium';
 type DropdownImageSize = 'small' | 'big';
 
 export type DropdownDataOption = {
-    label: string;
+    label: string
```
ffffffelix, line 11: You missed the semicolon

src/components/Dropdown/Dropdown.types.ts
```diff
@@ -30,6 +30,26 @@ export type DropdownDataTree = TreeViewData[]
 export type HandleSelect = (e: React.MouseEvent | React.KeyboardEvent, item?: DropdownDataOption) => void;
 
 export type DropdownProps = {
+
```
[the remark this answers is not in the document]
ffffffelix: After verification, we can also remove `multiple` and `isTree` from the props, as well as from the associated story.

PR #1164, opened 2025-10-22
fix: Clear dropdown stories for a better usage 

BouklifYacine:
Deleted those stories : 

- Playground
- Default 
- Default values 
- Outlined variant with search
- Outlined variant with big images
- Outlined variant with small images
- with Description
- With pill
- Empty 

Added stories  : 

For each new stories we have 4 examples of dropdown  ( with description, with small image, with big image and with Pill ) 

- Flat data 
- Flat data multiple 
- Grouped data 
- Grouped data multiple 
- Tree data 
- Tree data multiple 

Added more data for the new stories and updated images

### DIFF
--- a/src/components/Menu/Menu.spec.tsx
+++ b/src/components/Menu/Menu.spec.tsx
@@ -14,71 +14,135 @@ describe('Menu', () => {
         expect(queryByTestId('moonstone-menu')).not.toBeInTheDocument();
     });
 
-    it('should show the correct search results', async () => {
-        const user = userEvent.setup();
+    describe('search functionality', () => {
+        it('should show the correct search results', async () => {
+            const user = userEvent.setup();
 
-        render(
-            <Menu isDisplayed hasSearch data-testid="moonstone-menu">
-                <MenuItem label="Item1"/>
-                <MenuItem label="Item2"/>
-                <MenuItem label="Item3"/>
-            </Menu>
-        );
-        await user.type(screen.getByRole('searchbox'), 'item2');
+            render(
+                <Menu isDisplayed hasSearch data-testid="moonstone-menu">
+                    <MenuItem label="Item1"/>
+                    <MenuItem label="Item2"/>
+                    <MenuItem label="Item3"/>
+                </Menu>
+            );
+            await user.type(screen.getByRole('searchbox'), 'item2');
 
-        expect(screen.queryByText(/item1/i)).not.toBeInTheDocument();
-        expect(screen.getByText(/item2/i)).toBeInTheDocument();
-        expect(screen.queryByText(/item3/i)).not.toBeInTheDocument();
+            expect(screen.queryByText(/item1/i)).not.toBeInTheDocument();
+            expect(screen.getByText(/item2/i)).toBeInTheDocument();
+            expect(screen.queryByText(/item3/i)).not.toBeInTheDocument();
+        });
+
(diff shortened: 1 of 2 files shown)


BouklifYacine:
Could you check why the images on the multiple dropdown stories are not showing?

src/data/dropdownDataImages.tsx
```diff
@@ -1,69 +1,60 @@
...
+        image: <img src="https://picsum.photos/id/237/200/600"/>
+    },
+    {
+        label: 'option 4 (disabled)',
+        value: 'img-4',
+        isDisabled: true,
+        image: <img src="https://picsum.photos/id/237/200/300"/>
+    },
+    {
+        label: 'option 5 (disabled)',
```
ffffffelix: The label says `option 5 (disabled)`, but the item isn't disabled. Since we already have a disabled option, I guess you can remove (disabled) from this one.
```suggestion
        label: 'option 5',
```

src/data/dropdownDataTreeDescriptions.tsx
```diff
@@ -0,0 +1,125 @@
+import {TreeViewData} from '~/components/TreeView/TreeView.types';
+import {Folder, Lock} from '~/icons';
+
+export const dropdownDataTreeDescriptions: TreeViewData[] = [
```
ffffffelix: It seems we can't display a description with a tree. In that case, we don't need this data and the associate story

src/data/dropdownDataImages.tsx
```diff
@@ -1,69 +1,60 @@
...
-            },
-            {
-                label: 'option 3 with very long long label label label label label label label label',
-                value: '3',
-                image: <img src="https://via.placeholder.com/500?text=DropdownOptionImage"/>
-            }
-        ]
+        label: 'option 1',
+        value: 'img-1',
+        image: <img src="https://picsum.photos/id/237/200/300"/>
```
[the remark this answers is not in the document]
BouklifYacine: If i remove the id part there's no image displaying

src/data/dropdownDataGroupedDescriptions.tsx
```diff
@@ -0,0 +1,114 @@
+import {Folder, Lock} from '~/icons';
+
+export const dropdownGroupedDataDescriptions = [
+    {
+        groupLabel: 'test for item list with image',
```
ffffffelix: Rename the group label because we don't use image anymore. Propostion : `Group with description`

src/data/dropdownDataImages.tsx
```diff
@@ -1,69 +1,60 @@
-export const dropdownDataImages = [
+import {DropdownDataOption} from '~/components/Dropdown/Dropdown.types';
```
ffffffelix:
```suggestion
import type {DropdownDataOption} from '~/components/Dropdown/Dropdown.types';
```

src/data/dropdownDataGroupedDescriptions.tsx
```diff
@@ -0,0 +1,114 @@
+import {Folder, Lock} from '~/icons';
+
+export const dropdownGroupedDataDescriptions = [
```
ffffffelix: Finally, let's remove data with descriptions to avoid having too many datasets to maintain. So, we will add an example with a description in another dataset. (Sorry for the last-minute changes.)

PR #1168, opened 2025-10-28
chore: add vitest-addon for storybook

BouklifYacine:
- Add unit test in storybook for components


package.json
```diff
@@ -73,8 +73,10 @@
   "devDependencies": {
     "@jahia/eslint-config": "^2.1.2",
     "@jahia/stylelint-config": "^0.0.3",
+    "@mdx-js/react": "^3.1.1",
```
ffffffelix: Why do you need to add this dependency?

ffffffelix (changes requested):
I can't run accessibility tests from the Storybook UI, can you ?

PR #1183, opened 2025-11-18
feat: New DataTable component

### LINKED ISSUE #1153: New Table component based on TanStack Table
### Context
The current library ([react-table](https://www.npmjs.com/package/react-table)) used for our Table component is deprecated and has been replaced with [TanStack Table](https://tanstack.com/table/latest).

As our current Table component uses the library as a peer dependency, changing the library is considered a breaking change.

### Solution
Create a new Table component that uses TanStack Table without exposing the library to avoid this kind of issue in the future.

### Requirement
- The library must be a dependency of moonstone
- We must not expose the full API of the library.

### Implementation
- No regression, the component must have the same features (manage tree, and flat data, pagination)
- Must respond to future improvements (eg: stick first and last columns)
- Easy to use for developers
   - Create `TableCell` component with different variants to manage all type of data

### Documentation:
- [Comment](https://github.com/Jahia/moonstone/commit/407986a2c71a0bec148a40c693175bfe0400d3dd) from the first implementation
- Table usages in the product: ([GoogleDoc](https://docs.google.com/spreadsheets/d/1xGxKC_76q0XaBV-sjQ-h2QgAXmNTamSE11-im6HEAcM/edit?gid=1542960103#gid=1542960103))

### Related to
- https://github.com/Jahia/moonstone/issues/598
- https://github.com/Jahia/pm/issues/2325

BouklifYacine:
Created new custom types for the component’s columns and properties, independent from TanStack Table.

Refactored the main component to internally convert our custom column format into the format expected by TanStack.

Simplified Storybook stories to use the new abstracted API.

Removed utility functions and intermediate types that exposed TanStack-specific implementation details.


ffffffelix (commented):
As it's a POC will do not merge it right away and we will use this branch to finalize the component. For now, can you put this PR as draft, please ?

In order to simplify the code, I guess you can only keep one data set.

src/components/TablePoc/userTableConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const userTableConfig: TableColumnConfig<UserDataTypes>[] = [
+    {
+        key: 'firstName',
+        header: 'Prénom'
+    },
+    {
+        key: 'lastName',
+        header: 'Nom',
+        cell: value => (value as string).toUpperCase()
```
ffffffelix: For the first iteration, we will not change the value of the data. We will just display the value.

src/components/DataTable/types/MoonstoneTableColumn.types.ts
```diff
@@ -0,0 +1,12 @@
+import type {ReactNode} from 'react';
+
+export interface MoonstoneTableColumn<T extends Record<string, unknown>> {
+    key: keyof T;
+    label: string;
+    render?: (value: T[keyof T], row: T) => ReactNode;
+}
+
+export interface MoonstoneTableProps<T extends Record<string, unknown>> {
```
ffffffelix: Following my comment below, I prefer to name it `dataTableProps`.

src/components/DataTable/data/DataTable.tsx
```diff
@@ -0,0 +1,31 @@
+
+// The data to populate the table in Storybook
+
+import type {UserDataProps} from '../types/UserData.types';
```
ffffffelix: Since we have a simple dataset, TypeScript can infer the typing, so you don't have to specify it here. You can get rid of `UserData.types`

ffffffelix (changes requested):

package.json
```diff
@@ -101,6 +102,7 @@
     "rimraf": "6.1.2",
     "rollup-plugin-sbom": "^3.0.1",
     "sass": "^1.97.1",
+    "shx": "^0.4.0",
```
ffffffelix: Why has the lib been added?

src/components/DataTable/cells/TableCell.scss
```diff
@@ -0,0 +1,32 @@
...
+    &::-webkit-scrollbar-thumb {
+        background-color: var(--color-gray_light);
+    }
+}
+
+.moonstone-verticalAlignTop {
+    vertical-align: top;
+}
+
+.moonstone-verticalAlignCenter {
```
ffffffelix: I thought we should name this class `moonstone-verticalAlignMiddle`

src/components/DataTable/cells/TableStructuredCell.tsx
```diff
@@ -0,0 +1,70 @@
...
+            ...props
+        },
+        ref
+    ) => {
+        // Render nothing when no children are provided to avoid empty HTML markup
+        if (!children) {
+            return null;
+        }
+
+        const leftMarginIndentDepth = depth * 20; // Px
```
ffffffelix: Replace the `20` with a variable `indentSpace`.

src/components/DataTable/TableHead/TableHead.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+import type {TableHeadProps} from './TableHead.types';
+import './TableHead.scss';
+import clsx from 'clsx';
+
+export const TableHead: React.FC<TableHeadProps> = ({
+    isSticky = false,
```
ffffffelix: We will set `isSticky` to `true` by default

src/components/DataTable/TableBody/TableBody.tsx
```diff
@@ -0,0 +1,21 @@
...
+    className,
+    children,
+    ...props
+}) => React.createElement(
+    component,
+    {
+        className: clsx('moonstone-TableBody', className),
+        ...props
+    },
+    children
```
ffffffelix: We should return `null` when no children is provided

src/components/Table/table-cells/TableBodyCell.tsx
```diff
@@ -8,10 +8,10 @@ import {capitalize} from '~/utils/helpers';
 import {TableCell} from './TableCell';
 import './TableCell.scss';
 
-const TableBodyCellForwardRef: React.ForwardRefRenderFunction<HTMLDivElement, TableCellProps> = (
+const TableBodyCellForwardRef: React.ForwardRefRenderFunction<HTMLTableCellElement, TableCellProps> = (
     {
         component = 'td',
-        textAlign = 'left',
+        align = 'left',
```
ffffffelix: We must not modify the old `Table` component, and changing that prop can be considered a breaking change.

src/components/Table/table-cells/TableHeadCell.tsx
```diff
@@ -9,7 +9,7 @@ import {TableCell} from './TableCell';
 export const TableHeadCell: React.FC<TableCellProps> = ({
     component = 'th',
     width,
-    textAlign = 'left',
+    align = 'left',
```
ffffffelix: Same as my previous comment, we must not modify components related to the old Table.

ffffffelix (changes requested):

tsconfig.json
```diff
@@ -26,6 +33,13 @@
...
-}
+  "include": [
+    "src"
+  ],
+  "exclude": [
+    "node_modules",
+    "build",
+    "dist",
+    "types"
+  ]
```
ffffffelix: Why has the exclude configuration changed?

ffffffelix (changes requested):

src/components/DataTable/cells/TableHeadCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+            component="th"
+            width={width}
+            align={align}
+            verticalAlign={verticalAlign}
+            className={clsx('moonstone-TableHeadCell', className)}
+            aria-sort={ariaSort}
+            onClick={onClick}
+        >
+            <span className="flexRow_nowrap alignCenter">
+                {children}
```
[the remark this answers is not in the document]
BouklifYacine: When no children is provided, the span content is empty (React renders nothing for {children}). Returning null for the entire component would break table column alignment since <th> elements would be missing from the row.

ffffffelix (changes requested):

src/components/DataTable/cells/TableHeadCell.tsx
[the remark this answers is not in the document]
ffffffelix: Ok to keep it like that.

src/globals/_alignment.scss
```diff
@@ -9,3 +10,15 @@
 .textAlignRight {
     text-align: right;
 }
+
+.verticalAlignTop {
```
ffffffelix: Can you keep the `moonstone-` prefix, please? This helps prevent CSS collisions when we use our library.

ffffffelix (changes requested):

gflores-jahia (commented):
Looked ok from a high-level perspective. Really liked flexibility of customizing rendering of rows.

The big piece I see currently missing is option for uncontrolled state, or maybe event listener handlers (e.g. for restoring pagination, selection, expanded states tracked by redux for example). But this could be out of scope for this PR, or is a separate story.

Also agree with addition of virtualization as it seems virtualization seems to touch on a few interrelated components that can't be done with custom props.

ffffffelix:
@AKarmanov, @gflores-jahia 
We definitely plan to add the virtualization feature (https://github.com/Jahia/moonstone/issues/1204), but it will be added in a second batch of features.

I created a ticket to handle uncontrolled states: https://github.com/Jahia/moonstone/issues/1258

Thank you both for your feedback!

ffffffelix (approved):

PR #1192, opened 2025-11-21
feat: add checkbox features on the table component

BouklifYacine:
- Individual row selection: Users can select single rows using checkboxes in each row
- Multiple selection: A header checkbox allows users to select/deselect all rows at once
- Structured view compatibility: The checkbox functionality is fully compatible with the structured table view
- Add a visual sorting for the table working with the structured view and the multiple selection 
- Action column with empty label to show some button actions 
- Add default selection 
- Add Default sorting

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -31,7 +31,7 @@ const createTableColumns = <T extends Record<string, unknown>>(
                 return String(value ?? '');
             },
         meta: {
-            enableSorting: col.enableSorting
+            isSortable: col.isSortable
         }
     }));
 };
@@ -41,7 +41,7 @@ const adaptRowForTableBodyCell = <T extends Record<string, unknown>>(row: Row<T>
     isExpanded: row.getIsExpanded(),
     depth: row.depth,
     getToggleRowExpandedProps: () => ({
-        onClick: row.getToggleExpandedHandler(),
+        onClick: row.getToggleExpandedHandler()
     })
 });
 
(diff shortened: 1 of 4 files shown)


src/components/DataTable/DataTable.tsx
```diff
@@ -52,19 +54,26 @@ const extractIconFromCell = (cellValue: unknown): React.ReactElement | undefined
...
     const tableColumns = createTableColumns(columns);
 
     const table = useReactTable({
         data,
         columns: tableColumns,
+        state: {
+            expanded,
+            rowSelection
+        },
+        onExpandedChange: setExpanded,
```
ffffffelix: Why do you move `onExpandedChange` outside the `isStructured` condition?
It was a good idea to use it only when the data is structured.

src/components/DataTable/DataTable.tsx
```diff
@@ -98,7 +116,18 @@ export const MoonstoneTable = <T extends Record<string, unknown>>({
                     const adaptedRow = isStructured ? adaptRowForTableBodyCell(row) : undefined;
 
                     return (
-                        <TableRow key={row.id}>
+                        <TableRow
+                            key={row.id}
+                            isSelected={enableRowSelection ? row.getIsSelected() : undefined}
```
ffffffelix: I don't know if we really need the prop `isSelected` at the `TableRow` level. 
The only purpose of this prop is to add an unused CSS class (or at least in one test).

src/components/DataTable/DataTable.tsx
```diff
@@ -52,19 +54,26 @@ const extractIconFromCell = (cellValue: unknown): React.ReactElement | undefined
 export const MoonstoneTable = <T extends Record<string, unknown>>({
     data,
     columns,
-    isStructured = false
+    isStructured = false,
+    enableRowSelection = false
 }: DataTableProps<T>) => {
     const [expanded, setExpanded] = useState<ExpandedState>({});
+    const [rowSelection, setRowSelection] = useState<RowSelectionState>({});
```
ffffffelix: We also need to add a `defaultSelection` prop to automatically select rows when the table loads.

src/components/DataTable/TableData.stories.tsx
```diff
@@ -60,3 +61,74 @@ export const StructuredViewDataTable: Story = {
     name: 'Structured View'
 };
 
+export const SelectableDataTable: Story = {
+    render: (args: Omit<DataTableProps<UserDataRowProps>, 'data' | 'columns'>) => {
```
ffffffelix: I don't know why you omit `data` and `columns` for all stories?

src/components/DataTable/DataTable.tsx
```diff
@@ -52,18 +57,44 @@ const extractIconFromCell = (cellValue: unknown): React.ReactElement | undefined
...
+        setSortState(prevState => {
+            if (prevState?.columnId === columnId) {
+                if (prevState.direction === 'ascending') {
+                    return {columnId, direction: 'descending'};
+                }
+                return undefined;
+            }
+            return {columnId, direction: 'ascending'};
+        });
+    };
```
ffffffelix: If the component doesn't manage sorting itself, you shouldn't manage the direction or the sorted column, as they aren't reflected in the table data.

So we don't need a state. We just display the correctly sorted column in the right direction. You have to provide an `onClick` event on the `TableHead` so the consumer can manage it for us.

ffffffelix (approved):

PR #1208, opened 2025-12-02
feat: add solution for cell display in our new table component

BouklifYacine:
- Display the cells with our custom cells for specific case ( TableCellDate, TableCellNumber etc ) 
- For specific usecase the render solution he's still there to handle there own cell rendering 
- Can align a cell ( left, center, right )
- The cell hover actions is in progress 
- The cell status bar is in progress


BouklifYacine:
Closed PR to pull on the tanstack-table-poc branch instead of develop

PR #1221, opened 2025-12-17
feat: merged sortIndicator in tableheadcell component

BouklifYacine:
Merged sortIndicator in tableheadcell component

- Deleted sortindicator files to merge everything into tableheadcell files 
- Added test for tableheadcell and sortindicator 
- Enhanced DataTable to handle this change

### DIFF
--- a/src/components/DataTable/table-cells/TableHeadCell.spec.tsx
+++ b/src/components/DataTable/table-cells/TableHeadCell.spec.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import {render, screen, fireEvent} from '@testing-library/react';
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import {vi} from 'vitest';
 import {TableHeadCell} from './TableHeadCell';
 
@@ -32,7 +33,7 @@ describe('TableHeadCell', () => {
         expect(cell.tagName.toLowerCase()).toBe('th');
     });
 
-    it('should handle onClick on the th element', () => {
+    it('should handle onClick on the th element', async () => {
         const handleClick = vi.fn();
         render(
             <TableWrapper>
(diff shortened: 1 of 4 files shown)


ffffffelix (changes requested):

src/components/DataTable/table-cells/TableHeadCell.tsx
```diff
@@ -29,11 +35,17 @@ export const TableHeadCell: React.FC<TableHeadCellProps> = ({
...
             iconStart={iconStart}
             iconEnd={iconEnd}
-            style={style}
             onClick={onClick}
         >
-            {children}
-            {renderSortIcon({sortDirection, isSorted})}
+            <span className="moonstone-TableHeadCell-content flexRow_nowrap alignCenter">
+                {children}
+                {SortIcon && (
```
ffffffelix: - By default, we should define a default `sortDirection` value `descending`
- I guess we can use `isSorted` to determine whether the column is sorted.

src/components/DataTable/table-cells/TableHeadCell.tsx
```diff
@@ -29,11 +35,17 @@ export const TableHeadCell: React.FC<TableHeadCellProps> = ({
...
+                    <SortIcon
+                        aria-label={`Icon for sorting in ${sortDirection} order`}
```
ffffffelix: - We should avoid hard-coding `label` as our product is internationalized, even if it was in the previous `SortIndicator` component.
- Instead we can add [aria-sort](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort) at the `th` level to define which column is sorted and its direction

src/components/DataTable/table-cells/TableHeadCell.spec.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import {render, screen, fireEvent} from '@testing-library/react';
```
ffffffelix: As other tests, we use [userEvent](https://testing-library.com/docs/user-event/intro) instead of `fireEvent`

src/components/DataTable/table-cells/TableHeadCell.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+        });
+
+        it('should apply sorted active class when isSorted is true', () => {
+            render(
+                <TableWrapper>
+                    <TableHeadCell isSorted sortDirection="ascending" data-testid="cell"/>
+                </TableWrapper>
+            );
+            const sortIcon = screen.getByLabelText('Icon for sorting in ascending order');
+            expect(sortIcon).toHaveClass('moonstone-tableCellHead_sortActive');
```
ffffffelix: We should check `aria-sort` instead of checking CSS classes that can change.

ffffffelix (approved):

PR #1222, opened 2025-12-18
feat: Improved cell rendering into renderHelpers

BouklifYacine:
- Create renderHelpers.tsx with factory functions for cell rendering
- Rename DataTableColumnUtils.tsx to tableHelpers.tsx
- Add getCellValue (formerly getStringOrObjectValue) to tableHelpers
- Create utils/index.ts for barrel exports
- Merge stringColumn and stringOrObjectColumn into single stringColumn
- Remove duplicate tableCellUtils.tsx (unused)
- Add LocaleOptions type for Intl.NumberFormatOptions and Intl.DateTimeFormatOptions support
- Empty value handling centralized in TableCell ({children ?? '-'})
- Removed getcellvalue because renderString can do the same thing
- Added value params to render function helper to be able to use those function in tablecell for example
- RenderString handle only string for simplicity of usage

### DIFF
--- a/src/components/DataTable/utils/renderHelpers.tsx
+++ b/src/components/DataTable/utils/renderHelpers.tsx
@@ -1,23 +1,26 @@
 import React from 'react';
 import {Chip} from '~/index';
 
+type localeProp = string | string[];
+
 export type LocaleOptions = {
-    locale?: string | string[];
-    localeOptions?: Intl.NumberFormatOptions | Intl.DateTimeFormatOptions;
+    value: Date | null | undefined;
+    locale?: localeProp;
+    localeOptions?: Intl.DateTimeFormatOptions;
+} | {
+    value: number | null | undefined;
+    locale?: localeProp;
+    localeOptions?: Intl.NumberFormatOptions;
 };
 
 export const renderString = (value: string): React.ReactNode => value;
 
 export const renderNumber = (
-    value: number | null | undefined,
-    locale?: string | string[],
-    localeOptions?: Intl.NumberFormatOptions
+    {value, locale, localeOptions}: Extract<LocaleOptions, { value: number | null | undefined }>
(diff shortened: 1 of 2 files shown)


src/components/DataTable/utils/renderHelpers.tsx
```diff
@@ -0,0 +1,35 @@
...
+    locale?: string | string[];
+    numberOptions?: Intl.NumberFormatOptions;
+    dateOptions?: Intl.DateTimeFormatOptions;
+};
+
+export const renderString = () =>
+    (value: string | { value: string }): React.ReactNode =>
+        typeof value === 'string' ? value : value.value;
+
+export const renderNumber = (options?: LocaleOptions) =>
```
ffffffelix: All render functions should take a `value` parameter in order to be used in many contexts instead of being only usable in `columnHelpers` functions.

```
renderNumber(value, locale, localeOptions)
```

src/components/DataTable/utils/renderHelpers.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import {Chip} from '~/index';
+
+export type LocaleOptions = {
```
ffffffelix: As we change how the functions work, I guess we can also change the typing by `localeOptions = Intl.NumberFormatOptions | Intl.DateTimeFormatOptions`

src/components/DataTable/utils/tableHelpers.tsx
```diff
@@ -1,6 +1,14 @@
 import type {ColumnDef} from '@tanstack/react-table';
 import type {DataTableProps, SubRowKey} from '../DataTable.types';
 
+/**
+ * Extract string value from a field that can be string or { value: string }.
+ * Useful for sorting and comparison operations.
+ */
+export const getCellValue = (value: string | { value: string }): string => {
```
ffffffelix: Is this function still relevant?

ffffffelix (changes requested):

PR #1223, opened 2025-12-23
feat: reduce HTML nesting and improve TableCell architecture

BouklifYacine:
- Create TableStructuredCell for tree-view/expandable functionality
- Simplify TableCell to be a pure composable wrapper with children prop
- Remove IconTextIcon usage in table context (eliminates extra wrapper div)
- Replace TanStack Row prop with primitive props (depth, isExpandable, isExpanded, onToggleExpand)
- Delete unused TableCellChips, TableCellDate, TableCellNumber components
- Add flex layout with gap for cell content via CSS

### DIFF
--- a/src/components/DataTable/cells/TableCell.scss
+++ b/src/components/DataTable/cells/TableCell.scss
@@ -22,9 +22,9 @@
 }
 
 .moonstone-TableCell {
+    gap: var(--spacing-nano);
     padding-right: var(--spacing-medium);
     padding-left: var(--spacing-medium);
-    gap: var(--spacing-nano);
 
     overflow-y: auto;
 


src/components/DataTable/cells/TableCell.tsx
```diff
@@ -69,10 +33,12 @@ const TableCellForwardRef: React.ForwardRefRenderFunction<HTMLTableCellElement,
                 {flexFluid: typeof width === 'undefined'},
                 className
             )}
-            style={{width: width}}
+            style={{width}}
```
ffffffelix: Don't forget to add `props.style` to allow our user to add custom style here

src/components/DataTable/cells/TableStructuredCell.types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Whether this row can be expanded (has children)
+     */
+    isExpandable: boolean;
+
+    /**
+     * Whether this row is currently expanded
+     */
+    isExpanded: boolean;
```
ffffffelix: Should this prop be optional? Because if the item is not expandable there is no reason to provide `isExpanded`

src/components/DataTable/cells/TableStructuredCell.tsx
```diff
@@ -0,0 +1,83 @@
...
+                    'textAlign' + capitalize(textAlign),
+                    'moonstone-verticalAlign' + capitalize(verticalAlign),
+                    {flexFluid: typeof width === 'undefined'},
+                    className
+                )}
+                style={{width}}
+                {...props}
+            >
+                {renderContent()}
+            </Typography>
```
ffffffelix: Maybe we should `call `TableCell` directly instead of duplicating the code`?

src/components/DataTable/cells/TableCell.tsx
```diff
@@ -65,14 +30,17 @@ const TableCellForwardRef: React.ForwardRefRenderFunction<HTMLTableCellElement,
...
                 'textAlign' + capitalize(textAlign),
+                'flexRow',
+                'alignCenter',
                 'moonstone-verticalAlign' + capitalize(verticalAlign),
                 {flexFluid: typeof width === 'undefined'},
+                scrollableClass,
                 className
             )}
-            style={{width: width}}
+            style={{width, ...style}}
```
ffffffelix: I prefer to do it with `...props.style ` in order to avoid to declare another prop to just manage custom style

src/components/DataTable/cells/TableCell.tsx
```diff
@@ -29,16 +30,17 @@ const TableCellForwardRef: React.ForwardRefRenderFunction<HTMLTableCellElement,
             className={clsx(
                 'moonstone-TableCell',
                 'textAlign' + capitalize(textAlign),
+                'flexRow',
```
ffffffelix: I think we need to add `gap: var(--spacing-nano)` to `.moonstone-TableCell`

ffffffelix (approved):

PR #1226, opened 2026-01-02
feat: custom props to Table and show sort arrow on hover

BouklifYacine:


src/components/DataTable/table-cells/TableHeadCell.tsx
```diff
@@ -22,7 +22,8 @@ export const TableHeadCell: React.FC<TableHeadCellProps> = ({
 
     const sortClassName = isSortable && clsx(
         'moonstone-tableCellHead_sort',
-        {'moonstone-tableCellHead_sortActive': isActive}
+        {'moonstone-tableCellHead_sortActive': isActive},
+        {'moonstone-tableCellHead_sortHidden': !isActive}
```
ffffffelix: Instead of switching two classes, can we set it hidden by default and display it with `moonstone-tableCellHead_sortActive`

ffffffelix (changes requested):

PR #1229, opened 2026-01-06
refactor: refactor CardSelector thumbnail to use new Thumbnail component

BouklifYacine:


src/components/Thumbnail/Thumbnail.types.ts
```diff
@@ -0,0 +1,51 @@
...
+        /** Custom React element to render as thumbnail */
+        src: React.ReactElement;
+        alt?: never;
+        imgProps?: never;
+    }
+    | {
+        /** No src - renders fallback icon */
+        src?: undefined;
+        alt?: never;
+        imgProps?: never;
```
ffffffelix: Do we want to allow `imgProps` in that case to allow our user to add custom attribute to our thumbnail placeholder ?

src/components/CardSelector/CardSelector.types.ts
```diff
@@ -1,26 +1,22 @@
 import React from 'react';
 
-export type ThumbnailProps = {
+type BasicProps = Omit<React.ComponentPropsWithRef<'button'>, 'className' | 'id' | 'onClick'> & {
```
ffffffelix: Can we use the prop definition from the `Thumbnail` component to ensure consistent typing? Do we want to use the same prop name (`thumbnail`, `thumbnailAlt`) for both components?

ffffffelix (changes requested):

PR #1232, opened 2026-01-07
feat: Creating test for the new datatable component

BouklifYacine:


src/components/DataTable/cells/TableCell.spec.tsx
```diff
@@ -0,0 +1,150 @@
...
+        it('should render chevron right when expandable and not expanded', () => {
+            render(
+                <TableWrapper>
+                    <TableCell isExpandable depth={0} isExpanded={false}>Content</TableCell>
+                </TableWrapper>
+            );
+            // Search for chevron right icon (usually based on class or svg title if available)
+            // Moonstone icons usually have specific classes or we can check presence
+            // The code renders <ChevronRight className="moonstone-marginRightNano"/>
+            expect(document.querySelector('.moonstone-marginRightNano')).toBeInTheDocument();
```
ffffffelix: I guess it will be better to check `aria-expanded` instead of the presence of a class that it's not related to the expansion of the node?
We can also rewrite the description of the test to match what we test.

src/components/DataTable/TableRow/TableRow.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+    describe('State Classes', () => {
+        it('should apply isSelected class', () => {
+            render(
+                <TableWrapper>
+                    <TableRow isSelected data-testid="row">
+                        <td>Content</td>
+                    </TableRow>
+                </TableWrapper>
+            );
+            expect(screen.getByTestId('row')).toHaveClass('moonstone-TableRow-selected');
```
ffffffelix: It can be more relevant and semantic to test `aria-selected` instead of a specific class

src/components/DataTable/TableRow/TableRow.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+            <TableWrapper>
+                <TableRow ref={ref}>
+                    <td>Content</td>
+                </TableRow>
+            </TableWrapper>
+        );
+        expect(ref.current?.tagName).toBe('TR');
+    });
+
+    it('should forward className', () => {
```
ffffffelix: Change the label to `should add custom className`

src/components/DataTable/utils/utils.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+                // In de-DE, decimal separator is comma or different formatting.
+                // Simple checking presence of numbers
+                expect(result).not.toBeNull();
+            });
+        });
+
+        describe('renderDate', () => {
+            it('should format date', () => {
+                const date = new Date('2023-01-01T00:00:00Z');
+                expect(renderDate({value: date})).not.toBeNull();
```
ffffffelix: Same, we should apply some option to be sure the formatting is applied

src/components/DataTable/DataTable.spec.tsx
```diff
@@ -0,0 +1,264 @@
...
+                    data={data}
+                    columns={columns}
+                    primaryKey="id"
+                />
+            );
+
+            // Initial order: Alice (30), Bob (25), Charlie (35)
+            // Let's sort by Age
+            const ageHeader = screen.getByText('Age');
+            fireEvent.click(ageHeader);
```
ffffffelix: As with other tests I prefer to use `userEvent`

src/components/DataTable/DataTable.spec.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+            // Find checkbox for first row
+            // The first row of data is index 1 in role='row' (0 is header)
+            // But Checkbox component structure might vary.
+            // Rows have checkboxes.
+            const checkboxes = screen.getAllByRole('checkbox');
+            // Index 0 might be "Select All".
+
+            // Select first data row
+            await user.click(checkboxes[1]);
```
ffffffelix: From the previous comment, and the comment above can be removed.

```suggestion
            await user.click(firstRowCheckbox);
```

PR #1244, opened 2026-01-21
refactor: Manage no rendering for `Field` and `Fieldset`

BouklifYacine:
Return null when children/selector props are falsy to keep the DOM cleaner.

- Make `children` optional in Field and Fieldset components
- Make `selector` optional in FieldSelector component
- Add early return `null` for all three components when props are falsy


src/components/Field/FieldSelector/FieldSelector.types.ts
```diff
@@ -19,5 +19,5 @@ export type FieldSelectorProps = Omit<React.ComponentPropsWithRef<'div'>, 'class
     /**
      * FieldSelector selector
      */
-    selector: React.ReactElement;
+    selector?: React.ReactElement;
```
ffffffelix: We must keep `selector` as required

PR #1247, opened 2026-01-22
refactor: implement aliased exports for DataTable component

BouklifYacine:
- Adding the good export for DataTable component for avoiding conflict with Table component


src/components/DataTable/DataTable.tsx
```diff
@@ -17,11 +17,9 @@ import type {
 import {useState, useEffect, useMemo, useCallback} from 'react';
 
 import type {DataTableProps, CustomColumnMeta} from './DataTable.types';
-import {
-    Table,
-    TableRow,
-    Checkbox
-} from '~/index';
+import {Checkbox} from '~/index';
```
ffffffelix: As other components, we can import from `~/components`

PR #1256, opened 2026-01-27
chore: change stories format from csf 3 to csf next

### LINKED ISSUE #1211: Write stories with CSF next format
Storybook provides a new format for writing stories: [CSF next](https://storybook.js.org/docs/api/csf/csf-next)

### Tasks
- Convert all stories to the new CSF next format
- Use the [auto-migration script](https://storybook.js.org/docs/api/csf/csf-next#automatically) provided by Storybook
```
yarn dlx storybook automigrate csf-factories
```
- Check if there is no regression

BouklifYacine:


ffffffelix (changes requested):
- All docs pages are blank instead of displaying a default documentation page
- Initially, when we switched a component to `isReversed` (by setting the prop to `true` in the Control panel), the background of the story should change to a dark one to easily test the component in a dark context. Note that the implementation didn't work very well, and now it doesn't work at all.

PR #1273, opened 2026-02-04
feat: add a new column config for truncate/scroll option in all cells

BouklifYacine:
-Truncate/scroll applies to all cells: data cells, first column in structured view, selection checkbox column, and actions column
-First column in structured view now respects the column's truncate/scroll setting (same as other columns)
-Column config textOverflow ('truncate' or 'scroll') is supported; default is truncate
- Table layout is adjusted so column widths are respected and ellipsis/scroll work correctly in every cell
- Tests added/updated to cover fixed layout and truncate/scroll + canShrink on cells


src/components/DataTable/cells/TableCell.tsx
```diff
@@ -33,6 +34,9 @@ const TableCellForwardRef: React.ForwardRefRenderFunction<HTMLTableCellElement,
                 'alignCenter',
                 'verticalAlign' + capitalize(verticalAlign),
                 {flexFluid: typeof width === 'undefined'},
+                textOverflow === 'truncate' && 'moonstone-tableCell_truncate',
```
ffffffelix: The property `isNowrap` from the `Typography` component should already manage the truncate and set an ellipsis for long text.

src/utils/dataTable/tableHelpers.tsx
```diff
@@ -19,7 +19,8 @@ export const createTableColumns = <T extends Record<string, unknown>>(
         meta: {
             isSortable: col.isSortable ?? false,
             align: col.align ?? 'left',
-            width: col.width
+            width: col.width,
+            textOverflow: col.textOverflow ?? 'truncate'
```
ffffffelix: Should we need to set the default value `truncate` as we already set a default value in the component `TableCell` ?

src/components/DataTable/cells/TableCell.scss
```diff
@@ -18,3 +20,24 @@
...
     }
 }
+
+.moonstone-tableCell.moonstone-tableCell_truncate {
+    display: block;
+}
+
+.moonstone-tableCell_scroll {
+    display: block;
+    overflow-x: auto;
```
ffffffelix: I don't thing we must set `overflow-x: auto` as we don't want a vertical scrollbar in a cell.

ffffffelix (changes requested):
When I scroll into a cell the scrollbar overlap the text. It will be good to take a look to the old Table to see how we did.

PR #1278, opened 2026-02-10
feat: truncate cells by default and allow scrollbar in cell

### LINKED ISSUE #1271: Manage truncate and scrollable cell
We currently have two ways to manage too-long text in a cell:
- Truncate the text
- Add a horizontal scrollbar

### Tasks
Provide a way to manage these two behaviors with the column configuration.
The default behavior is to truncate the cell

BouklifYacine:
- Added overflow detection to DataTable cells (only show scrollbar on hover when content is truncated)
- Fixed Firefox cell vertical shift on hover
- Fixed Chrome scroll position reset bug
- Updated tests to use render() for components returned by render helpers in utils spec file


src/components/DataTable/cells/TableCell.scss
```diff
@@ -4,17 +4,67 @@
     gap: var(--spacing-nano);
     padding-right: var(--spacing-medium);
     padding-left: var(--spacing-medium);
+    overflow: hidden;
+}
+
+// Structured cell wrapper (expandable rows with chevron)
+.moonstone-tableCellExpandable {
```
ffffffelix: All className should follow the pattern `.moontone-${componentName}_${modificateur}`

src/components/DataTable/cells/TableCell.scss
```diff
@@ -4,17 +4,67 @@
...
+    flex: 1;
+    min-width: 0; // Allow shrinking below content width
+}
+
+.moonstone-tableCellExpandable__chevron {
+    flex-shrink: 0;
+}
```
ffffffelix: - Should follow the naming convention
- Do we really need it ? Can we use a utils class ?
- If not, it can be a good idea create a new class `.moonstone-tableStructuredCell_chevron` to add this `flex-shrink: 0` and also styles from `.moonstone-marginRightNano`

src/components/DataTable/cells/TableCell.scss
```diff
@@ -4,17 +4,67 @@
...
+// Structured cell wrapper (non-expandable rows with indentation only)
+.moonstone-tableCellIndent {
+    flex: 1;
+    min-width: 0;
+}
 
-    overflow-y: auto;
+.moonstone-cellText {
+    // WebKit-specific scrollbar customization (Chrome, Safari, Edge)
+    @supports selector(::-webkit-scrollbar) {
```
ffffffelix: Good idea to provide `@support` ! 👍

ffffffelix (changes requested):
Unfortunately, it doesn't work very well on Mac (Chrome / Firefox). I think the trick is to give the cell more space so the scrollbar appears.
- Set the height on the cell instead of the row
- The cell should take the full available height

ffffffelix (changes requested):

ffffffelix (changes requested):

ffffffelix (dismissed):

src/data/dataTable/TableData.tsx
```diff
@@ -2,17 +2,17 @@ import type {DataUser} from './dataColumnsUser';
...
 export const dataTable: DataUser[] = [
     {
-        firstName: 'Walter',
-        lastName: 'White',
+        firstName: 'Walter Hartwell White the Great Heisenberg',
+        lastName: 'White-McFarlane-Johnson-Williams',
         age: 52,
         status: 'Active',
-        progress: 99,
+        progress: 9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999,
```
GauBen, line 9: This will render as "1e+100", making it shorter than what you expect

src/data/dataTable/TableData.tsx
```diff
@@ -47,8 +47,8 @@ export const dataTable: DataUser[] = [
         ]
     },
     {
-        firstName: 'Jon',
-        lastName: 'Snow',
+        firstName: 'Jon Aegon Targaryen Snow the King in the North',
```
GauBen, line 50: SPOILER ALERT

src/utils/dataTable/renderHelpers.tsx
```diff
@@ -12,12 +13,45 @@ export type LocaleOptions = {
...
 
 export const renderNumber = (
     {value, locale, localeOptions}: Extract<LocaleOptions, { value: number | null | undefined }>
-): React.ReactNode => (value === null || value === undefined) ? null : value.toLocaleString(locale, localeOptions);
+): React.ReactNode => {
+    if (value === null || value === undefined) {
+        return null;
+    }
+
+    if (typeof value !== 'number') {
```
GauBen: While you're at it, you might want to support BigInts (e.g. `99...99n` <- `n` suffix for bigints)

src/utils/dataTable/renderHelpers.tsx
```diff
@@ -12,12 +13,45 @@ export type LocaleOptions = {
...
-): React.ReactNode => (value === null || value === undefined) ? null : value.toLocaleDateString(locale, localeOptions);
+): React.ReactNode => {
+    if (value === null || value === undefined) {
+        return null;
+    }
+
+    if (!(value instanceof Date)) {
+        console.warn('renderDate expects a Date object, received:', typeof value, value);
+        return null;
+    }
```
GauBen, line 50: While you're at it you might want to add support for Temporal, which reached Stage 4!

(Use `globalThis.Temporal?.(...)` because Safari does not support it yet, so it might be undefined 😮‍💨)

GauBen (dismissed):

src/data/dataTable/TableData.tsx
ffffffelix: In fact it's render as `10 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000 000`

src/utils/dataTable/renderHelpers.tsx
ffffffelix: Done !

src/utils/dataTable/renderHelpers.tsx
ffffffelix: Maybe not this time :'(

GauBen (approved):

PR #1281, opened 2026-02-18
fix: tablecells aligned with the tableheader on a flat structure

BouklifYacine:


src/components/DataTable/cells/TableStructuredCell.tsx
```diff
@@ -28,41 +26,30 @@ export const TableStructuredCell = React.forwardRef<HTMLTableCellElement, TableS
...
                 {...props}
             >
-                {renderContent()}
+                {isExpandable ? (
+                    <span
+                        className="moonstone-tableCellExpandable flexRow_nowrap alignCenter"
+                        style={{marginLeft: indent}}
+                        onClick={onToggleExpand}
+                    >
+                        {isExpanded ? <ChevronDown style={{marginRight: 0}}/> : <ChevronRight style={{marginRight: 0}}/>}
```
ffffffelix: - Why do you add `style={{marginRight: 0}}` ?
- It's better to manage style by adding a class than by using inline styling.

ffffffelix (changes requested):

ffffffelix (dismissed):

PR #1283, opened 2026-02-18
feat: move actions to row level with enableActions prop

BouklifYacine:
### Changes

- **Row-level actions**: Actions are injected via `renderRow` + `defaultRender({ actions })` or via the `renderActions` shorthand
- **enableActions**: New table-level prop to toggle the actions column (replaces legacy `actions` prop)
- **actionsLabel**: New optional prop for the column header label (empty by default)
- **renderActions**: Shorthand for defining per-row actions without custom row rendering
- **Hover visibility**: Actions are hidden by default and shown on row hover using the `moonstone-tableCell_actions` class
- **Compatibility**: `defaultRender` now accepts `{ actions?: ReactNode }` to inject custom actions per row

### Breaking changes

- `actions` prop is replaced by `enableActions` + `renderActions`
- `actionsHeaderLabel` prop is replaced by `actionsLabel` (empty by default)

### DIFF
--- a/src/components/DataTable/DataTable.spec.tsx
+++ b/src/components/DataTable/DataTable.spec.tsx
@@ -1,8 +1,8 @@
-import { render, screen, waitFor, within } from '@testing-library/react';
+import {render, screen, waitFor, within} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { describe, expect, it, vi } from 'vitest';
-import { DataTable, TableRow } from '~/components/DataTable';
-import { numberColumn, stringColumn } from '~/utils/dataTable';
+import {describe, expect, it, vi} from 'vitest';
+import {DataTable, TableRow} from '~/components/DataTable';
+import {numberColumn, stringColumn} from '~/utils/dataTable';
 
 type TestData = {
     id: string;
@@ -99,7 +99,7 @@ describe('DataTable', () => {
             />
         );
 
-        const buttons = screen.getAllByTestId('button-actions'); 
+        const buttons = screen.getAllByTestId('button-actions');
         expect(buttons).toHaveLength(3);
         buttons.forEach(button => {
             expect(button).toBeVisible();
@@ -122,7 +122,7 @@ describe('DataTable', () => {
             />
         );
 
-        const buttons = screen.getAllByTestId('button-hidden'); 
+        const buttons = screen.getAllByTestId('button-hidden');
         expect(buttons).toHaveLength(3);
         buttons.forEach(button => {
             expect(button).not.toBeVisible();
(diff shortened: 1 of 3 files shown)


src/components/DataTable/cells/TableCell.scss
```diff
@@ -18,3 +18,11 @@
         background-color: var(--color-gray_light);
     }
 }
+
+.moonstone-tableCell_actions {
+    visibility: hidden;
+
+    tr:hover & {
```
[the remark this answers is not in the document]
ffffffelix: I am agree to this comment

src/components/DataTable/DataTable.types.ts
```diff
@@ -174,28 +174,42 @@ type SelectionProps =
...
 
-    /**
-     * The label for the actions column header
-     */
-    actionsHeaderLabel?: string;
-};
+          /**
+           * Header label for the actions column. Empty by default; pass a string when a visible label is needed.
+           */
+          actionsLabel?: string;
```
ffffffelix: For consistency, I suggest not allowing any label for the action column header. You can remove this prop and its associated code.

src/components/DataTable/DataTable.spec.tsx
```diff
@@ -60,19 +60,58 @@ describe('DataTable', () => {
...
+    it('should render actions column when enableActions and renderActions are provided', () => {
         render(
             <DataTable<TestData>
+                enableActions
                 data={data}
                 columns={columns}
                 primaryKey="id"
-                actions={row => <button type="button">Edit {row.name}</button>}
+                actionsLabel="Actions"
+                renderActions={row => <button type="button">Edit {row.name}</button>}
```
ffffffelix: I would like to remove the ability to manage actions at the `DataTable` level and allow them to be managed only through the `defaultRender` function.

src/components/DataTable/cells/TableCellActions/TableCellActions.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const TableCellActions: React.FC<TableCellActionsProps> = ({
+    displayMode = 'hover',
+    children
+}) => (
+    <TableCell
+        align="right"
+        className={
+            displayMode === 'hover' ?
+                'moonstone-tableCell_actions moonstone-tableCell_actions--hover' :
+                'moonstone-tableCell_actions'
```
ffffffelix: We should be able to display the action on hover over the row **and** always show it. With this implementation, we have two choices between two behaviors but cannot have both at the same time.

ffffffelix (changes requested):

ffffffelix (changes requested):

src/components/DataTable/DataTable.tsx
```diff
@@ -127,10 +126,12 @@ export const DataTable = <T extends NonNullable<unknown>>({
         }
     }, [data, isStructured, table]);
 
+    const hasActionsColumn = Boolean(renderRow);
```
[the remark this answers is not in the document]
ffffffelix: This flag is not relevant. See comment above.

PR #1284, opened 2026-02-19
feat: add resize column for the datatable 

### LINKED ISSUE #1198: Manage resize columns
Allow the user to manage each column's size

### Tasks
- Add a feature tag `enableResize` to allow the user to resize columns
- On `hover` a cell, we will display the vertical border that trigger the column resizing
- Each column can have a `min-width`, a `max-width`, and a default `width`. By default, we can provide values for each of these props.
- Add `onResizeStart` callback when the resize begins
- Add `onResizeStop` callback when the resize is finished
- Add `onResizing` callback when resizing

BouklifYacine:


src/components/DataTable/cells/TableHeadCell.scss
```diff
@@ -15,5 +15,36 @@
...
+        transition: opacity 0.15s ease, border-color 0.15s ease;
+
+        user-select: none;
+        touch-action: none;
+    }
+
+    /* More specific selector first for no-descending-specificity */
+    &.moonstone-tableHeadCell_resizing .moonstone-tableHeadCell_resizeHandle,
+    &:hover .moonstone-tableHeadCell_resizeHandle {
+        border-left-color: var(--color-gray);
```
ffffffelix: When an element is active you can use `$dnd-highlight: var(--color-accent_light60);` from `globals/_dnd.scss`

src/components/DataTable/DataTable.types.ts
```diff
@@ -99,6 +117,50 @@ export type DataTableBaseProps<T extends NonNullable<unknown>> = {
...
+    /**
+     * Callback when column sizing changes (controlled mode). Use with columnSizing for persistence.
+     */
+    onColumnSizingChange?: OnChangeFn<ColumnSizingState>;
+
+    /**
+     * Width of the actions column in pixels when enableResize is true.
+     * @default 60
+     */
+    actionsColumnWidth?: number | string;
```
ffffffelix: - I don't think we should set a width to the action column, otherwise we can have truncation or a scrollbar in that column 😱
- We should unify the typings of width props `number` or `string` but not both

src/components/DataTable/cells/TableHeadCell.tsx
```diff
@@ -12,6 +12,12 @@ export const TableHeadCell = ({
     children,
     sorting,
     onClick,
+    enableResize,
+    resizeHandler,
+    onResizeReset,
```
ffffffelix: For the first iteration we should not allow any reset, so you can remove `onResizeReset`

src/components/DataTable/cells/TableHeadCell.tsx
```diff
@@ -46,6 +58,17 @@ export const TableHeadCell = ({
                     />
                 )}
             </span>
+            {enableResize && resizeHandler && (
+                <div
+                    data-resizer
+                    className="moonstone-tableHeadCell_resizeHandle"
+                    role="separator"
+                    aria-label={resizeHandleLabel}
```
ffffffelix: I bet this prop will never be used, but in order to provide a way to do it, I suggest having a prop `resizeHandleAttr` to allow our user to put any atrubute to this element. It also can be good for tests.

ffffffelix (changes requested):
In UX point of view:
- I will prefer to not always displays "resizeHandler" but only display it on hover of the handler.
- I would like to allow to resize cells from any row and not only from then header

ffffffelix (changes requested):
In UX point of view:
- I will prefer to not always displays "resizeHandler" but only display it on hover of the handler.
- I would like to allow to resize cells from any row and not only from then header

PR #1312, opened 2026-03-24
feat: add the the resize column for datatable 

### LINKED ISSUE #1198: Manage resize columns

BouklifYacine:
## Description

Description of the Pull Request

## Tests

- [ ] Unit Tests
- [ ] Accessibility is OK

## Checklist

- [ ] Example in storybook
- [ ] Allows custom props

### DIFF
--- a/src/components/DataTable/DataTable.spec.tsx
+++ b/src/components/DataTable/DataTable.spec.tsx
@@ -1,4 +1,4 @@
-import {render, screen, waitFor, within} from '@testing-library/react';
+import {fireEvent, render, screen, waitFor, within} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {describe, expect, it, vi} from 'vitest';
 import {DataTable, TableRow} from '~/components/DataTable';
@@ -530,6 +530,35 @@ describe('DataTable pagination feature', () => {
         await waitFor(() => {
             expect(screen.queryByText('Alice')).not.toBeInTheDocument();
             expect(screen.getByText('Bob')).toBeInTheDocument();
+        });
+    });
+});
+
+describe('DataTable resize feature', () => {
+    it('should call resize callbacks when a column handle is dragged', async () => {
+        const onResizeStart = vi.fn();
+        const onResizeStop = vi.fn();
+
+        render(
+            <DataTable<TestData>
+                enableResize
+                data={data}
+                columns={columns}
+                primaryKey="id"
+                onResizeStart={onResizeStart}
+                onResizeStop={onResizeStop}
+            />
+        );
+
+        const firstHeaderResizeHandle = document.querySelector('th .moonstone-tableCell_resizeHandle');
+        expect(firstHeaderResizeHandle).toBeInTheDocument();
+
+        fireEvent.mouseDown(firstHeaderResizeHandle as Element, {clientX: 200});
+        fireEvent.mouseUp(document, {clientX: 200});
+
+        await waitFor(() => {
+            expect(onResizeStart).toHaveBeenCalled();
+            expect(onResizeStop).toHaveBeenCalled();
         });
     });
 });
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -8,9 +8,8 @@ import {
 } from '@tanstack/react-table';
 
 import type {ExpandedState, Row} from '@tanstack/react-table';
-import {useState, useEffect, useMemo, useCallback} from 'react';
-
-import {useTableSelection, useTableSorting, useTablePagination} from '~/hooks';
+import {useState, useEffect, useMemo, useCallback, useRef} from 'react';
+import {useTableSelection, useTableSorting, useTablePagination, useTableResizing} from '~/hooks';
 import type {DataTableProps, CustomColumnMeta, DefaultRenderOptions} from './DataTable.types';
 import {Checkbox} from '~/components';
 import {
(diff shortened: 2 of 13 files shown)


src/components/DataTable/DataTable.spec.tsx
```diff
@@ -1,4 +1,4 @@
-import {render, screen, waitFor, within} from '@testing-library/react';
+import {fireEvent, render, screen, waitFor, within} from '@testing-library/react';
```
ffffffelix, line 1: Use userEvent

src/components/DataTable/DataTable.spec.tsx
```diff
@@ -533,3 +533,32 @@ describe('DataTable pagination feature', () => {
...
+                enableResize
+                data={data}
+                columns={columns}
+                primaryKey="id"
+                onResizeStart={onResizeStart}
+                onResizeStop={onResizeStop}
+            />
+        );
+
+        const firstHeaderResizeHandle = document.querySelector('th .moonstone-tableCell_resizeHandle');
```
ffffffelix, line 553: Avoid using css classes as selector for test, use role or attribute instead

src/components/DataTable/DataTable.tsx
```diff
@@ -206,7 +261,11 @@ export const DataTable = <T extends NonNullable<unknown>>({
 
     return (
         <>
-            <Table className={className} {...props}>
+            <Table
+                className={className}
+                {...props}
+                style={props.style}
```
ffffffelix, line 267: Why do you need to add style. It will be spread throw the props

src/components/DataTable/DataTable.types.ts
```diff
@@ -12,6 +12,8 @@ export type CustomColumnMeta = {
     isSortable?: boolean;
     align?: 'left' | 'center' | 'right';
     width?: string;
+    minWidth?: number;
+    maxWidth?: number;
```
ffffffelix, line 16: Why `minWidth` and `maxWidth` are number whereas `width` is string ?
We should have the same type for similar props

src/components/DataTable/cells/TableCell.scss
```diff
@@ -2,6 +2,33 @@
 
 $cell-height: 48px;
 
+.moonstone-tableCell--resizable {
```
ffffffelix, line 5: Please can you follow the naming convention.

src/components/DataTable/cells/TableCell.scss
```diff
@@ -2,6 +2,33 @@
...
+
+    user-select: none;
+    touch-action: none; // Let TanStack handle touch events instead of the browser
+
+    &:hover {
+        background-color: var(--color-accent);
+    }
+}
+
+.moonstone-tableCell_resizeHandle--active {
```
ffffffelix, line 28: Please follow the naming convention

src/components/DataTable/cells/TableCell.scss
```diff
@@ -2,6 +2,33 @@
...
+    right: 0;
+    bottom: 0;
+
+    width: 2px;
+
+    background-color: var(--color-gray_light);
+    cursor: col-resize;
+
+    user-select: none;
+    touch-action: none; // Let TanStack handle touch events instead of the browser
```
ffffffelix, line 21: We don't care about touch action as our application is not compatible to touch screen

src/components/DataTable/cells/TableCell.tsx
```diff
@@ -44,6 +48,18 @@ const TableCellForwardRef: React.ForwardRefRenderFunction<HTMLTableCellElement,
             {...props}
         >
             {children ?? '-'}
+            {resizeHandler && (
+                <div
+                    className={clsx(
+                        'moonstone-tableCell_resizeHandle',
+                        isResizing && 'moonstone-tableCell_resizeHandle--active'
```
ffffffelix, line 55: Do you check if there is a semantic attribute or aria attribute that we can add ?

src/components/DataTable/DataTable.types.ts
```diff
@@ -69,6 +71,23 @@ export type DataTableColumn<T extends NonNullable<unknown>> = {
      * When undefined, the column takes all available space.
      */
     width?: string;
+
+    /**
+     * Whether this column can be resized when table-level resizing is enabled.
+     * Requires a `width` in pixels to be set — without it, the column has no known initial
+     * size and may jump when the user starts dragging. Defaults to true.
+     */
+    enableResizing?: boolean;
```
ffffffelix, line 80: It can be confusing to have to set resizing at two levels (table and each cells).
Maybe we should consider applying `enableResizing` at the cell level, in that case we can guess if the resizing feature is enabled at least one cell is resizable.

src/data/dataTable/dataColumnsUser.tsx
```diff
@@ -20,6 +20,7 @@ export const dataColumnsUser: DataTableColumn<DataUser>[] = [
     {
         key: 'firstName',
         label: 'User',
+        minWidth: 120, // Can help to prevent columns from being too narrow during resizing
```
ffffffelix, line 23: We need to set a default `minWidth` and ` maxWidth` for our preset render.

ffffffelix (changes requested):

PR #1314, opened 2026-03-27
feat: manage status cell

ffffffelix:
## Description
- Change how we managed to display cells before and after they rendered based on the data
- Create a new component `TableCellStatus` to handle statuses in the table
- Change the default value for the number of items allowed per page

## Tests

- [ X ] Unit Tests
- [ ] Accessibility is OK


src/components/DataTable/TableRow/TableRow.scss
```diff
@@ -1,43 +1,54 @@
...
     .moonstone-chip {
         color: var(--color-light);
 
         background-color: var(--color-accent_light40);
     }
 
-    .moonstone-button:not(:hover, :focus) {
-        color: var(--color-light);
-    }
+    // .moonstone-button:not(:hover, :focus) {
```
BouklifYacine, line 49: Delete if unused

src/components/DataTable/TableRow/TableRow.scss
```diff
@@ -1,43 +1,54 @@
...
+    //     }
 
-        &.hasFocus {
-            border: 2px solid var(--color-accent_dark_contrast);
-        }
-    }
+    //     &.hasFocus {
+    //         border: 2px solid var(--color-accent_dark_contrast);
+    //     }
+    // }
```
BouklifYacine, line 41: Same, delete those classes if they’re not used.

src/components/DataTable/DataTable.tsx
```diff
@@ -61,6 +67,58 @@ export const DataTable = <T extends NonNullable<unknown>>({
...
+        // when the data or primary key changes.
+        if (headerCellWidths.current) {
+            headerCellWidths.current = {};
+        }
+    }, [data, primaryKey]);
+
+    // Sync pending flags (written during render) to state after each render.
+    // useLayoutEffect fires synchronously before the browser paints, so the
+    // follow-up re-render that adds the custom columns is invisible to the user.
+    useLayoutEffect(() => {
```
BouklifYacine: Can you explain what is this hook and if we really need it?

src/components/DataTable/DataTable.stories.tsx
```diff
@@ -212,25 +211,29 @@ export const AllFeaturesTable: Story = {
...
+                                    </>
+                                </TableCellStatus>
                             ),
-                            actionsOnHover: (
-                                <>
-                                    <Button icon={<Visibility/>} variant="ghost"/>
-                                    <Button icon={<Edit/>} variant="ghost"/>
-                                    <Button icon={<Delete/>} variant="ghost"/>
-                                </>
+                            after: (
```
BouklifYacine, line 224: I like the idea of having a before and after " slots " for our cells. 

In the future we could also check if we can be able to have multiple cells to put in after or before slots

src/components/DataTable/DataTable.tsx
```diff
@@ -19,13 +19,18 @@ import {
...
     TableCell,
-    TableCellActions,
     TableStructuredCell,
     TableHeadCell
 } from '~/components/DataTable';
 import {createTableColumns} from '~/utils/dataTable/tableHelpers';
 import {Pagination} from '~/components/Pagination';
 
+// Custom column metadata type
+type CustomColumnType = 'before' | 'after';
```
BouklifYacine: Move this type in a .types file

src/components/DataTable/DataTable.tsx
```diff
@@ -61,6 +67,58 @@ export const DataTable = <T extends NonNullable<unknown>>({
...
+    const pendingCustomAfter = useRef(false);
+
+    // Store custom cell content
+    const customCellsMap = useRef<Map<string, {
+        before?: React.ReactNode;
+        after?: React.ReactNode;
+    }>>(new Map());
+
+    // Clear custom cell mappings when the underlying data or primary key changes
+    useEffect(() => {
```
BouklifYacine, line 86: Do we need UseEffect?

src/components/DataTable/DataTable.tsx
```diff
@@ -85,7 +143,53 @@ export const DataTable = <T extends NonNullable<unknown>>({
...
     });
 
-    const tableColumns = useMemo(() => createTableColumns(columns), [columns]);
+    // Build columns with optional display columns for custom before/after cells
+    const tableColumns = useMemo(() => {
+        const baseColumns = createTableColumns(columns);
+        const finalColumns = [];
+
+        // Helper to create a custom display column with width measurement
+        const createCustomColumn = (position: 'before' | 'after', index: number) => ({
```
BouklifYacine: You have to check if its the reason the cells in a larger screen are not aligned with the header 

Image

BouklifYacine (changes requested):
- The "Before" and "After" idea is good — I like it — but maybe we could simplify the code, because I found it a bit difficult to follow. That said, this is just a suggestion.
- Try using useEffect instead of useLayoutEffect (as React recommends). It may impact performance if not implemented properly (see link below).
- The DataTable is not aligned with the header cells on large screens or when the window is resized. This might be caused by a CSS conflict or by the before/after implementation (see screenshot below).
- The status bar does not fully cover the cell content. For example, if the first cell contains long text, the last letter remains visible (see screenshot below).

https://fr.react.dev/reference/react/useLayoutEffect


Image
Image

src/components/DataTable/TableRow/TableRow.scss
ffffffelix: As we don't manage the highlighted state yet, I commented out the corresponding style. But we will need it soon.

src/components/DataTable/DataTable.tsx
ffffffelix: When the user adds a custom cell before/after, we need to calculate the width to insert the corresponding TableCellHead. This `useEffect` is to manage when the data changes (filtering, new columns ...)

src/components/DataTable/DataTable.tsx
ffffffelix: Because we only calculate the cell's width at the first rendering, and not when the component is resized. I am taking a look.

src/components/DataTable/DataTable.tsx
ffffffelix: As we calculate the width of the TableCellHead to add when the user uses custom cells throw before/after, I used `useLayoutEffect` to perform the calculation before rendering.

src/components/DataTable/DataTable.stories.tsx
ffffffelix: Good idea, I add the management as several cells throw before / after.

src/components/DataTable/DataTable.tsx
ffffffelix: It should be fixed

BouklifYacine (approved):

PR #1332, opened 2026-04-10
fix: import DataTable and related utils

ffffffelix:
## Description
- Fix DataTable import
- Allow to import DataTable/utils
- Fix the typing of `renderHelpers.tx`

PR #1335, opened 2026-04-17
refactor: Datatable improve component organisation 

### LINKED ISSUE #1316: Organize DataTable components
As the component is growing quickly, we need to keep it well-organized.

### Tasks
- Each sub-component should be in a dedicated folder within `DataTable` (eg: Table)
- Each cell-related component should be in a dedicated folder within `DataTable/cells` (eg: TableCell, TableStructuredCell)
- Extract the logic and type of the pagination component and remove it from the datatable for a better structure 

### Questions
- Should we move `TableHeadCell` to a folder within `DataTable` or keep it under the `cells` folder?
- Can we split DataTable files by the features function and only keep the rendering?
- Can we split typing in `DataTable.types` to avoid having all types in one files

BouklifYacine:
- Refactored DataTable internals into dedicated subfolders without breaking public exports.
- Move exposed helpers into DataTable/shared (renderHelpers, columnHelpers, tableHelpers).
- Keep internal rendering helpers under DataTable/utils .


src/components/DataTable/cells/TableCellActions/TableCellActions.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import clsx from 'clsx';
-import {TableCell} from '../TableCell';
+import {TableCell} from '../TableCell/TableCell';
```
ffffffelix: As you created an index file, you can import from 
`import {TableCell} from '../index';`
or with alias `import {TableCell} from '~/components/DataTable/cells';`

This comment can be applied to all imports.

src/components/DataTable/cells/index.ts
```diff
@@ -1,11 +1,11 @@
-export * from './TableCell';
-export * from './TableCell.types';
+export * from './TableCell/TableCell';
```
ffffffelix: Can you also add an `index` for all subfolders (TableCell, TableCellActions, ...) in order to simplify these import files.

src/components/DataTable/Columns/dataTableColumns.tsx
```diff
@@ -0,0 +1,83 @@
...
+import {TableCell} from '../cells/TableCell/TableCell';
+import {TableHeadCell} from '../cells/TableHeadCell/TableHeadCell';
+import {TableStructuredCell} from '../cells/TableStructuredCell/TableStructuredCell';
+import type {
+    CustomColumnMeta,
+    RenderDataTableBodyCellsProps,
+    RenderDataTableHeaderCellsProps
+} from './dataTableColumns.types';
+
+export const renderDataTableHeaderCells = <T extends NonNullable<unknown>>({
```
ffffffelix: I would prefer to name it `renderHeadCell` to match the component the function returns.

src/components/DataTable/Columns/dataTableColumns.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+                onClickTableHeadCell?.(header.id);
+            }}
+            >
+                {flexRender(header.column.columnDef.header, header.getContext())}
+            </TableHeadCell>
+        );
+    });
+
+export const renderDataTableBodyCells = <T extends NonNullable<unknown>>({
```
ffffffelix: I would prefer to name it `renderCell` to match the component that the function returns.

src/components/DataTable/utils/renderCells.types.ts
```diff
@@ -0,0 +1,22 @@
+import type {HeaderGroup, Row} from '@tanstack/react-table';
+import type {DataTableProps} from '../DataTable.types';
+
+export type CustomColumnMeta = {
+    isSortable?: boolean;
+    align?: 'left' | 'center' | 'right';
+    width?: string;
+    isScrollable?: boolean;
+    cellProps?: React.TdHTMLAttributes<HTMLTableCellElement> & Record<string, unknown>;
+};
```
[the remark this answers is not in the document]
ffffffelix: Right !

src/components/DataTable/hooks/dataTableCustomCells.types.ts
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import type {DataTableProps} from '../DataTable.types';
+
+export type DataTableCustomCellPosition = 'before' | 'after';
```
ffffffelix: Maybe we can go with something shorter without losing too much information, like : `customCellPosition` as we are already in a DataTable folder.

src/components/DataTable/hooks/dataTableCustomCells.types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type DataTableCustomHeaderWidths = Record<DataTableCustomCellPosition, string[]>;
```
ffffffelix:
```suggestion
export type customHeaderWidths = Record<DataTableCustomCellPosition, string[]>;
```

src/components/DataTable/hooks/dataTableCustomCells.types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type UseDataTableCustomCellsProps<T extends NonNullable<unknown>> = {
```
ffffffelix:
```suggestion
export type useCustomCellsProps<T extends NonNullable<unknown>> = {
```

src/components/DataTable/hooks/dataTableCustomCells.types.ts
```diff
@@ -0,0 +1,25 @@
...
+    data: T[];
+    primaryKey: string;
+    renderRow?: DataTableProps<T>['renderRow'];
+};
+
+export type UseDataTableCustomCellsReturn = {
```
ffffffelix:
```suggestion
export type UseCustomCellsReturn = {
```

src/components/DataTable/hooks/useDataTableCustomCells.ts
```diff
@@ -0,0 +1,118 @@
+import React, {useCallback, useEffect, useRef, useState} from 'react';
+import type {
+    DataTableCustomHeaderWidths,
+    UseDataTableCustomCellsProps,
+    UseDataTableCustomCellsReturn
+} from './dataTableCustomCells.types';
+
+const createInitialHeaderWidths = (): DataTableCustomHeaderWidths => ({before: [], after: []});
```
ffffffelix: In order to match other props when we set a state with a default value:
```suggestion
const defaultHeaderWidths = (): DataTableCustomHeaderWidths => ({before: [], after: []});
```

src/components/DataTable/hooks/useDataTableCustomCells.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export function useDataTableCustomCells<T extends NonNullable<unknown>>({
```
ffffffelix:
```suggestion
export function UseCustomCells<T extends NonNullable<unknown>>({
```

src/components/DataTable/pagination/dataTablePagination.types.ts
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import type {Table as TanstackTable} from '@tanstack/react-table';
+import type {PaginationProps as ComponentPaginationProps} from '~/components/Pagination';
+
+type PaginationBaseProps = {
```
ffffffelix, line 4: I am wondering whether all these type definitions shouldn't be in the `Pagination` component instead.

ffffffelix (changes requested):
- Can you also wrap the component `Table` into a folder
- Add an index for each sub-folder in order to simplify some imports
- We have to find a naming convention for those functions:
	- `renderFoo`
	- `getBar`
	- `useMyHook`
- Some folders are capitalized, while others are not. I suggest capitalizing only the component folder name.
- I am wondering whether it would be better to add a `hooks` folder inside `DataTable`, since those functions are very specific to the DataTable component and will not be used elsewhere. We will keep the root `hooks` folder for functions shared across all components. In that case, we may remove the `columns` and `customCells` folders and move the corresponding functions into `DataTable/hooks`

This PR can be impacted by : https://github.com/Jahia/moonstone/pull/1332

src/components/DataTable/hooks/useDataTableCustomCells.ts
ffffffelix: Can you also change the function name, and also the file name ?

src/components/DataTable/hooks/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './useDataTableCustomCells';
+export * from './useTablePagination';
+export * from './useTableSelection';
+export * from './useTableSorting';
```
ffffffelix: We should have a consistent naming for all these hooks.
As those hooks are already imported from `DataTable/hooks` I suggest naming them:
- `useCustomCells`
- `usePagination`
- `useSelection`
- `useSorting`

In any case, we should have a consistent name between the file and the function name

ffffffelix (changes requested):
I don't like the `columns/dataTableColumns` because it's not suited to the functions used in `renderHeadCell` and `renderCells`.
I am also wondering whether it is still relevant to keep the function in `utils/DataTable`, or if we should group all functions related to DataTable into its own folder.

We just have to be careful about functions that are only for internal use and those that are exposed for external use.

src/components/DataTable/pagination/dataTablePagination.ts
```diff
@@ -0,0 +1,22 @@
+import type {PaginationProps as ComponentPaginationProps} from '~/components/Pagination';
+import type {DataTablePaginationLogicProps} from './index';
+
+export const getPaginationProps = <T extends NonNullable<unknown>>({
```
ffffffelix: From my point of view, this function is unnecessary because it only maps props, and we use it only once.
So I suggest removing this function and implementing it where we use it.

ffffffelix (changes requested):
Finally, I suggest creating a `DataTable/shared` folder to manage all functions and helpers we want to expose outside the library, and a `DataTable/utils` folder for functions used internally only.

Check the function in the shared folder is exposed, and we can close the PR #1332

src/components/index.ts
```diff
@@ -1,4 +1,5 @@
 export {DataTable} from './DataTable';
+export * from './DataTable/shared';
```
ffffffelix: I am not sure we want to add it.
All shared DataTable resourcesmust only be accessible from `@jahia/moonstone/Datatable`

src/components/DataTable/hooks/useCustomCells.types.ts
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
ffffffelix, line 1: We wanted to rename this file `useCustomCells.types.ts` ?

ffffffelix (changes requested):

ffffffelix (dismissed):

ffffffelix (approved):

PR #1341, opened 2026-04-27
ci: deploy storybook only on release

### LINKED ISSUE #1331: Only deploy Storybook when we release
Currently, we always deploy Storybook when we merge to main.

As we plan to use Storybook as our documentation platform, we must deploy only when we release to ensure documentation that matches the current version.

We also get this warning during the [publication to gh-pages](https://github.com/Jahia/moonstone/actions/runs/24987502375) : 
```
[deploy-storybook](https://github.com/Jahia/moonstone/actions/runs/24987502375/job/73164730708#step:11:2)
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: peaceiris/actions-gh-pages@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

### Tasks
- Only deploy Storybook to gh-pages when we release
- Fix the warning during Storybook CI

BouklifYacine:
- Deploy storybook only when there are a new release


.github/workflows/release.yml
```diff
@@ -4,10 +4,15 @@ on:
...
       - main
+  release:
+    types:
+      - published
+      - released
   workflow_dispatch: null
 
 jobs:
   publish-release:
+    if: ${{ github.event_name != 'release' }}
```
ffffffelix: I don't think we should change anything in the `publish-release` workflow, but only change when the `deploy-storybook` is triggered : on release.

ffffffelix (changes requested):

.github/workflows/release.yml
```diff
@@ -28,20 +32,31 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   deploy-storybook:
+    # only run on non-prerelease GitHub releases
+    if: ${{ github.event_name == 'release' && !github.event.release.prerelease }}
```
GauBen: You need to use job output to know if a release was created

https://github.com/GauBen/chachalog/releases/tag/chachalog%400.5.1

.github/workflows/release.yml
```diff
@@ -23,25 +25,36 @@ jobs:
...
+      - id: chachalog
+        run: npx chachalog@^0.5.1 publish-release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   deploy-storybook:
-    runs-on: ubuntu-latest
     needs: publish-release
+    # only run when chachalog released @jahia/moonstone
+    if: ${{ needs.publish-release.outputs.released_moonstone == 'true' }}
```
GauBen:
```suggestion
    if: needs.publish-release.outputs.released_moonstone
```

I'll change `true` to a real version number at some point

GauBen (dismissed):

GauBen (approved):

ffffffelix (approved):

PR #1354, opened 2026-04-30
feat: use css-module for Datetime component

### LINKED ISSUE #1348: Add CSS modules for Datetime component

With our migration to CSS Modules, the upcoming Date Selector needs to be built using CSS Modules.

The components that need to be migrated to CSS Modules are:

- DatetimeInput
- TimeInput

BouklifYacine:
- Migration of datetime component in css module


src/components/Input/DateTimeInput/DateTimeInput.tsx
```diff
@@ -79,6 +79,7 @@ export const DateTimeInput = React.forwardRef<HTMLInputElement, DateTimeInputPro
     const selectedDate = sanitizedValue.date;
     const [isCalendarOpen, setIsCalendarOpen] = useState(false);
     const [displayedMonth, setDisplayedMonth] = useState(sanitizedValue.date || new Date());
+    const lastSelectedDateTimestampRef = useRef(selectedDate?.getTime() ?? null);
```
ffffffelix, line 93: I guess this PR was only about CSS modules ;)

src/components/Input/DateTimeInput/DateTimeInput.tsx
```diff
@@ -103,13 +116,13 @@ export const DateTimeInput = React.forwardRef<HTMLInputElement, DateTimeInputPro
     };
 
     return (
-        <div className={clsx('moonstone-dateTimeInput', className)}>
+        <div className={clsx(['moonstone-dateTimeInput', styles['moonstone-dateTimeInput']], className)}>
```
ffffffelix: As this is a new component, we don't need to keep the classic classes.
The classic classes are kept for other components for compatibility purposes to prevent breaking changes for people who rely on them.
For new components, we go with full CSS modules.

src/components/Input/DateTimeInput/DateTimeInput.tsx
```diff
@@ -103,13 +116,13 @@ export const DateTimeInput = React.forwardRef<HTMLInputElement, DateTimeInputPro
...
             <BaseInput
                 ref={ref}
                 {...props}
                 readOnly
                 containerRef={calendarAnchorRef}
-                className="moonstone-dateTimeInput_dateField"
+                className={clsx('moonstone-dateTimeInput_dateField', styles['moonstone-dateTimeInput_dateField'])}
```
ffffffelix: As we don't keep old classes, you don't need to prefix CSS classes with `.moonstone-${componentName)` for new components. You can go with shorter classes like : `dateField` as the classe is scoped for the component context.

src/components/Input/TimeInput/TimeInput.tsx
```diff
@@ -93,11 +93,11 @@ export const TimeInput = React.forwardRef<HTMLInputElement, TimeInputProps>(({
     };
 
     return (
-        <div className={clsx('moonstone-timeInput', layout.flexRow_nowrap, layout.alignCenter, className)}>
+        <div className={clsx(['moonstone-timeInput', styles['moonstone-timeInput']], layout.flexRow_nowrap, layout.alignCenter, className)}>
```
ffffffelix: I think you must import the layout.
`import {layout} from '~/globals/css-utils.js';`

ffffffelix (changes requested):

src/components/Input/DateTimeInput/DateTimeInput.tsx
BouklifYacine: I know but in the build i got some errors in the test

ffffffelix (changes requested):
You also have to add CSS import in `src/globals.scss`

ffffffelix (approved):

PR #1361, opened 2026-05-15
fix: styling issues due to CSS specificity

### LINKED ISSUE #2323: Styling issues after moonstone v2.20 upgrade
### Environment and versions used

Jahia 8.2.3.0
jContent 3.7.0-SNAPSHOT

### Steps to reproduce

**Hover button**
1. Go to jContent
2. Hover some buttons

Image

**Disabled dropdown** in Content Editor
1. Install QA-modules
2. Add content `dependentChoicelist`

Image

### Current Behavior

The issue is due to [CSS Modules](https://github.com/Jahia/moonstone/issues/1250) in Moonstone. We get CSS specificity/override issues, so let's have the same kind of side effect.
The Dropdown issue has been reported by @gflores-jahia in that comment: https://github.com/Jahia/jcontent/pull/2259#issuecomment-4425466889

### Desired Behavior

**Hover button**
Image

**Disabled Dropdown**
Image

The fix will be in Moonstone, but I created the ticket to track the issue at the product level.

ffffffelix:
## Description
- Improve the visual tests jobs to allow to downloads baseline and current screenshots related to failed jobs
- Update default color use in storybook to `--moon-color-gray_dark`
- Use `@mixin` instead of scss placeholder to define variant styles
- Refactor button styles to have a better split between all possibilities (`variant`, `color`, `size`), also, we now have dedicated attributes as selectors. So I removed old classes (e.g `moonstone-button_ghost_accent`)
- The component `ButtonToggle` uses the component `Button` under the hood instead of duplicating the button component.
- Fix the disabled style of the Dropdown component when it's used as ghost variant by displaying a background
- Unify some styles for consistency
- I updated the visual tests
- I removed reset at the component level because it complexifies CSS overrides, so now we do a really light reset on each component.


BouklifYacine (dismissed):

gflores-jahia (dismissed):

src/components/ButtonToggle/ButtonToggle.module.scss
```diff
@@ -11,47 +9,25 @@
...
-    &:hover:not(.moonstone-button_loading, [disabled]) {
-        color: var(--moon-color-accent);
-
-        background-color: var(--moon-color-gray_light20);
-    }
 }
 
-.moonstone-button.moonstone-buttonToggle_pressed {
+// Sorry about this uggly selector, but it's needed to override the defaut button styles.
+.moonstone-buttonToggle.moonstone-buttonToggle.moonstone-buttonToggle_pressed {
```
gflores-jahia: from my (limited) css understanding, it looks like we are overriding some definitions of color for pressed state. As alternative, I'm wondering if possible to also do this through vars i.e. define default styling vars and then be able to set them here if it needs to be overridden. (But I don't know if specificity is still an issue to redefine those vars).

gflores-jahia (dismissed):
lgtm. I like the minimal reset per component. 

I did add some minor comment around specificity for consideration.

src/globals/reset.module.scss
GauBen, line 1: 🎉

GauBen (dismissed):
godspeed

GauBen (dismissed):

src/components/ButtonToggle/ButtonToggle.module.scss
ffffffelix: I finally changed the selector to a nested selector. Thanks for the highlight.

GauBen (approved):

PR #1370, opened 2026-05-26
feat: add new drawer component

BouklifYacine:
- Add a new drawer component

### DIFF
--- a/.chachalog/Fb83UAkp.md
+++ b/.chachalog/Fb83UAkp.md
@@ -3,4 +3,4 @@
 "@jahia/moonstone": minor
 ---
 
-Add new drawer component (#1370)
+Add new drawer component (Beta) (#1370)
--- a/.chachalog/Fb83UAkp.md
+++ b/.chachalog/Fb83UAkp.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Add new drawer component (#1370)
--- a/src/components/Drawer/Drawer.module.scss
+++ b/src/components/Drawer/Drawer.module.scss
@@ -1,0 +1,43 @@
+.moonstone-drawer_overlay {
+    display: flex;
+    justify-content: flex-end;
+
+    background: var(--moon-color-dark60);
+}
+
+.moonstone-drawer {
+    position: relative;
+
+    display: flex;
+    flex-direction: column;
+
+    height: 100vh;
+    overflow-y: auto;
+
+    background-color: var(--moon-color-white);
+    box-shadow: -2px 0 8px 0 rgba(0, 0, 0, 0.15);
+
+    animation: moonstone-drawer_slideIn 0.25s ease-out;
+}
+
+.moonstone-drawer_small {
+    width: 300px;
+}
+
+.moonstone-drawer_medium {
+    width: 450px;
+}
+
+.moonstone-drawer_large {
+    width: 640px;
+}
+
+@keyframes moonstone-drawer_slideIn {
+    from {
+        transform: translateX(100%);
+    }
+
+    to {
+        transform: translateX(0);
+    }
+}
--- a/src/components/Drawer/Drawer.module.scss
+++ b/src/components/Drawer/Drawer.module.scss
@@ -1,43 +1,8 @@
-.moonstone-drawer_overlay {
-    display: flex;
-    justify-content: flex-end;
-
-    background: var(--moon-color-dark60);
-}
-
-.moonstone-drawer {
-    position: relative;
-
+.drawer {
     display: flex;
     flex-direction: column;
 
-    height: 100vh;
+    height: stretch;
+    min-height: 0;
     overflow-y: auto;
-
-    background-color: var(--moon-color-white);
-    box-shadow: -2px 0 8px 0 rgba(0, 0, 0, 0.15);
-
-    animation: moonstone-drawer_slideIn 0.25s ease-out;
-}
-
-.moonstone-drawer_small {
-    width: 300px;
-}
-
-.moonstone-drawer_medium {
-    width: 450px;
-}
-
-.moonstone-drawer_large {
-    width: 640px;
-}
-
-@keyframes moonstone-drawer_slideIn {
-    from {
-        transform: translateX(100%);
-    }
-
-    to {
-        transform: translateX(0);
-    }
 }
--- a/src/components/Drawer/Drawer.spec.tsx
+++ b/src/components/Drawer/Drawer.spec.tsx
@@ -1,0 +1,39 @@
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {Drawer} from './index';
+
+describe('Drawer', () => {
+    it('should display content when open', () => {
+        render(<Drawer isOpen data-testid="moonstone-drawer">Drawer content</Drawer>);
+        expect(screen.getByText('Drawer content')).toBeInTheDocument();
+    });
+
+    it('should not display content when closed', () => {
+        render(<Drawer isOpen={false} data-testid="moonstone-drawer">Drawer content</Drawer>);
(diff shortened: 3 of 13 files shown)


src/components/Drawer/Drawer.module.scss
```diff
@@ -0,0 +1,43 @@
+.moonstone-drawer_overlay {
+    display: flex;
+    justify-content: flex-end;
+
+    background: var(--moon-color-dark60);
+}
```
ffffffelix: You don't need any overlay.

src/components/Drawer/Drawer.module.scss
```diff
@@ -0,0 +1,43 @@
...
+    display: flex;
+    flex-direction: column;
+
+    height: 100vh;
+    overflow-y: auto;
+
+    background-color: var(--moon-color-white);
+    box-shadow: -2px 0 8px 0 rgba(0, 0, 0, 0.15);
+
+    animation: moonstone-drawer_slideIn 0.25s ease-out;
```
ffffffelix: Your current animation isn't very smooth and seems buggy.

src/components/Drawer/Drawer.types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Callback fired when the open state changes (e.g. ESC key, click outside)
+     */
+    onOpenChange?: (open: boolean) => void;
+
+    /**
+     * Size of the Drawer
+     */
+    size?: 'small' | 'medium' | 'large';
```
ffffffelix: I don't know yet whether we need any custom size. So for now, I prefer not to specify any size.

src/components/Drawer/Drawer.types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Whether the Drawer is open
+     */
+    isOpen: boolean;
+
+    /**
+     * Callback fired when the open state changes (e.g. ESC key, click outside)
+     */
+    onOpenChange?: (open: boolean) => void;
```
ffffffelix: I am feeling confused with the description and the examples.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+        role,
+        dismiss
+    ]);
+
+    return (
+        children && isOpen && (
+            <FloatingPortal>
+                <FloatingOverlay lockScroll className={clsx('moonstone-drawer_overlay', styles['moonstone-drawer_overlay'])}>
+                    <FloatingFocusManager context={context}>
+                        <div
```
ffffffelix: I think we should use the `Paper` component as a wrapper for the `Drawer`. it also avoid you to add custom style.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    const {getFloatingProps} = useInteractions([
+        role,
+        dismiss
+    ]);
+
+    return (
+        children && isOpen && (
+            <FloatingPortal>
+                <FloatingOverlay lockScroll className={clsx('moonstone-drawer_overlay', styles['moonstone-drawer_overlay'])}>
```
ffffffelix: For now, I plan to have the drawer displayed aside the content and not above, so I guess we don't need floating-ui for the moment.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    <FloatingFocusManager context={context}>
+                        <div
+                            ref={drawerRef}
+                            className={clsx(
+                                reset,
+                                ['moonstone-drawer', styles['moonstone-drawer']],
+                                [`moonstone-drawer_${size}`, styles[`moonstone-drawer_${size}`]],
```
ffffffelix: You don't need to add good old CSS classes for new components. You can go for CSS modules only.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+                                className
+                            )}
+                            aria-modal="true"
```
ffffffelix: As the component will be displayed alongside the content, I am not sure if `aria-modal` will still be relevant.

src/components/Drawer/Drawer.module.scss
```diff
@@ -0,0 +1,43 @@
+.moonstone-drawer_overlay {
```
ffffffelix: As we go for CSS modules only you don't need to prefix the class with `moonstone-`

src/components/Drawer/Drawer.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+    });
+
+    it('should add extra classname', () => {
+        render(<Drawer isOpen data-testid="moonstone-drawer" className="extra">Drawer content</Drawer>);
+        expect(screen.getByTestId('moonstone-drawer')).toHaveClass('extra');
+    });
+
+    it('should apply size modifier class', () => {
+        render(<Drawer isOpen data-testid="moonstone-drawer" size="large">Drawer content</Drawer>);
+        expect(screen.getByTestId('moonstone-drawer')).toHaveClass('moonstone-drawer_large');
```
ffffffelix: As the component will become a full CSS module, you cannot rely on CSS classes for tests.

src/components/Drawer/Drawer.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {Drawer} from './index';
+
+describe('Drawer', () => {
```
ffffffelix, line 4: Maybe we can also check if the Drawer is hide by default, what do you think ?

src/components/Drawer/Drawer.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+        const user = userEvent.setup();
+
+        render(
+            <Drawer isOpen data-testid="moonstone-drawer" onOpenChange={handleOpenChange}>
+                Drawer content
+            </Drawer>
+        );
+
+        await user.keyboard('{Escape}');
+        expect(handleOpenChange).toHaveBeenCalledWith(false, expect.anything(), expect.anything());
```
ffffffelix: So the function `onOpenChange` doesn't fire when the component is hidden/displayed?

src/components/Drawer/Drawer.module.scss
```diff
@@ -0,0 +1,43 @@
...
+    background: var(--moon-color-dark60);
+}
+
+.moonstone-drawer {
+    position: relative;
+
+    display: flex;
+    flex-direction: column;
+
+    height: 100vh;
```
ffffffelix: I don't think we should set the height to `100vh`. The Drawer will take the full height of the available space, but not the full height of the browser.

src/components/Drawer/Drawer.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {Button, Typography} from '~/components';
+
+const meta = {
+    title: 'Components/Drawer',
+    component: Drawer,
+    tags: ['beta'],
+    args: {
+        isOpen: false,
+        children: null as DrawerProps['children']
+    },
```
ffffffelix: Do you have to set `args`? Can we use the default value of the props defined in the component?

ffffffelix (changes requested):

.chachalog/Fb83UAkp.md
```diff
@@ -0,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Add new drawer component (#1370)
```
ffffffelix:
```suggestion
Add new drawer component (Beta) (#1370)
```

ffffffelix (changes requested):
You also have to add the style import of the component in `src/globals.scss` when moonstone is not used in "scope" mode.

ffffffelix (approved):

PR #1376, opened 2026-06-04
fix: Fix repositioning on context menu

gflores-jahia:
## Description

Fix an issue where menu (In this case context menu) loses anchor when menu resizes.

We now add a `ResizeObserver` so that position is recomputed when size of menu changes.

- Refactored existing useEffect and created a `computePosition` method to be able to reuse in the resize observer as well.
- Added test for usePositioning hook, covering resize scenario

## Tests

- [x] Unit Tests
- [ ] Accessibility is OK

## Checklist

- [ ] Example in storybook
- [ ] Allows custom props

### DIFF
--- a/.chachalog/1P6lxSvc.md
+++ b/.chachalog/1P6lxSvc.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": patch
+---
+
+Fix context menu repositioning on resize (#1376)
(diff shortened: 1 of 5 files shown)


ffffffelix (dismissed):

gflores-jahia:
had to resolve conflict and fix some lint issues (but wasn't part of my PR).

ffffffelix (approved):

PR #1378, opened 2026-06-09
feat: Add drawer slot to LayoutContent

BouklifYacine:
- Add a new slot for the Drawer component in LayoutContent.


src/layouts/content/LayoutContent.module.scss
```diff
@@ -16,3 +20,7 @@
     max-width: 800px;
     margin: 0 auto;
 }
+
+.moonstone-layoutContent_storyDrawer {
```
ffffffelix: I don't want a CSS class specific to a story into the styles of the component.

src/layouts/content/LayoutContent.spec.tsx
```diff
@@ -16,6 +16,13 @@ describe('LayoutContent', () => {
         expect(screen.getByText(requiredProps.children)).toBeInTheDocument();
     });
 
+    it('should display drawer alongside content', () => {
+        const {container} = render(<LayoutContent {...requiredProps} drawer={<aside>test-drawer</aside>}/>);
+        expect(screen.getByText(requiredProps.children)).toBeInTheDocument();
+        expect(screen.getByText('test-drawer')).toBeInTheDocument();
+        expect(container.querySelector('.moonstone-layoutContent_body')).toBeInTheDocument();
```
ffffffelix: - Why do you test the presence of `requiredProps.children` in a test about the drawer?
- Why do you test if `.moonstone-layoutContent_body` is here?
- Don't query classes anymore; we should avoid it as we plan to move to full CSS modules

src/layouts/content/LayoutContent.stories.tsx
```diff
@@ -49,6 +53,18 @@ export const WithoutPadding: Story = {
     }
 };
 
+export const WithDrawer: Story = {
+    render: args => (
+        <LayoutContent
+            header={<Header title="Header"/>}
+            {...args}
+            drawer={<Drawer isOpen className={styles['moonstone-layoutContent_storyDrawer']}>Drawer content</Drawer>}
```
ffffffelix: Finally, I hesitate to use a prop in `LayoutContent`. Maybe it would be better to let the consumer display a Drawer near the children content?

src/layouts/content/LayoutContent.tsx
```diff
@@ -35,25 +53,17 @@ export const LayoutContent = React.forwardRef(({
...
-                        ) :
-                        children ?? content
-                      )
-                }
-            </div>
+            {drawer ? (
+                <div className={clsx(
+                    ['flexRow_nowrap', layout.flexRow_nowrap],
+                    ['flexFluid', layout.flexFluid],
+                    ['moonstone-layoutContent_body', styles['moonstone-layoutContent_body']]
```
ffffffelix: If it's a new CSS class, you should use CSS module only.

src/layouts/content/LayoutContent.stories.tsx
BouklifYacine: If we just say "put the Drawer next to the content" without a dedicated prop, it ends up stuck in the same zone as the content — so it gets the padding and scrolls with it, which breaks its fixed side-panel look.

By keeping a dedicated drawer prop, LayoutContent handles the separation between the two zones correctly. Otherwise, every team that wants a drawer would have to redo this work themselves.

So for me we keep the drawer prop — it's simpler and more consistent for everyone.

ffffffelix:
Finally, we don't need it. We can reopen it if we change our minds.

PR #1382, opened 2026-06-10
feat: add new utils class to manage layout

BouklifYacine:
- Adds new utility classes for recurring flex + gap patterns while keeping existing layout helpers unchanged.


BouklifYacine:
_This PR does not change `flexCol_center` because `flexCol_center` and `alignCenter` have different responsibilities, and merging them would be a breaking change._

.chachalog/qmQR0EdD.md
```diff
@@ -0,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Add row/col gap utility classes (#1382)
```
ffffffelix: We don't need a changelog as this class will be only for internal purposes.

ffffffelix (commented):
I don't want to expose those classes for now. Can you check that they're not exposed in the built CSS.

src/utils/_layout-internal.module.scss
```diff
@@ -0,0 +1,39 @@
+@use '../tokens/spacings/map' as *;
```
ffffffelix, line 1: Styles inside `global` are exposed in the global style.
I prefer to keep those classes internal for now. I suggest to put this file into `src/utils` instead of the `globals` folder.

src/utils/_layout-internal.module.scss
```diff
@@ -0,0 +1,39 @@
...
+%flex-row {
+    display: flex;
+    flex-flow: row wrap;
+    min-height: 0;
+}
+
+// ---
+// Gap utilities
+// ---
+@each $name, $value in $spacings {
```
ffffffelix, line 18: I think these classes should have `flex-wrap: nowrap` by default.

src/components/GlobalStyle/GlobalStyle.stories.tsx
```diff
@@ -133,3 +133,40 @@ export const Direction = () => (
 export const Justify = () => <>{displayItems('row', 'justify')}</>;
 
 export const Alignment = () => <>{displayItems('row', 'align')}</>;
+
+export const GapUtilities = () => (
```
ffffffelix: As I want to keep these classes internal for now. We don't need to document them in storybook.

ffffffelix (changes requested):

PR #1385, opened 2026-06-16
fix: improve controlled DataTable features

ffffffelix:
## Description
- **Breaking change**: Introduce a new row context API so row and cell callbacks use stable data instead of exposing TanStack objects.
- Improve controlled DataTable behavior for sorting, pagination, and row expansion to keep state handling consistent and predictable.
- Allow to control which node is expanded by default when the Table is Structured
- Add support for dynamic row and cell props (including custom attributes) for better per-row customization.
- Prevent focus styles when a row is clicked.

## Tests

- [x] Unit Tests
- [ ] Accessibility is OK

### DIFF
--- a/.chachalog/lNFAy4J-.md
+++ b/.chachalog/lNFAy4J-.md
@@ -2,4 +2,6 @@
 "@jahia/moonstone": patch
 ---
 
-Improved DataTable controlled sorting, pagination, and row expansion behavior
+Changed DataTable row and cell callbacks to use a stable context object instead of internal table rows, and improved controlled sorting, pagination, and row expansion behavior.
+
+If you customize DataTable row or cell rendering, update your callbacks to use the new `id`, `data`, `meta`, and `render` fields.
--- a/.chachalog/lNFAy4J-.md
+++ b/.chachalog/lNFAy4J-.md
@@ -2,4 +2,4 @@
 "@jahia/moonstone": patch
 ---
 
-Improve controlled features of DataTable
+Improved DataTable controlled sorting, pagination, and row expansion behavior
--- a/.chachalog/lNFAy4J-.md
+++ b/.chachalog/lNFAy4J-.md
@@ -1,0 +1,5 @@
+---
+"@jahia/moonstone": patch
+---
+
+Improve controlled features of DataTable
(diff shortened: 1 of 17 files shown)


.chachalog/lNFAy4J-.md
```diff
@@ -0,0 +1,7 @@
+---
+"@jahia/moonstone": patch
+---
+
+Changed DataTable row and cell callbacks to use a stable context object instead of internal table rows, and improved controlled sorting, pagination, and row expansion behavior.
+
+If you customize DataTable row or cell rendering, update your callbacks to use the new `id`, `data`, `meta`, and `render` fields.
```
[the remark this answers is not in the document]
ffffffelix: As the component is in Beta, we can do breaking changes

src/components/DataTable/hooks/useExpansion.ts
```diff
@@ -0,0 +1,50 @@
...
+import type {ExpandedState} from '@tanstack/react-table';
+
+type UseExpansionProps = {
+    expandedRows?: string[];
+    defaultExpandedRows?: true | string[];
+    onExpandChange?: (expandedRows: string[]) => void;
+};
+
+function toExpandedState(ids: string[]): ExpandedState {
+    return ids.reduce<Record<string, boolean>>((acc, id) => ({...acc, [id]: true}), {});
```
GauBen: Easier to read

```suggestion
    return Object.fromEntries(ids.map(id => [id, true]))
```

src/components/DataTable/shared/columnHelpers.tsx
```diff
@@ -21,21 +21,21 @@ type ColumnOptions = Omit<LocaleOptions, 'value'> & {
  * ];
  */
 export const stringColumn = <T, >(get: (row: T) => string, options?: ColumnOptions) => ({
-    render: (value: string) => renderString(value),
+    render: ({data}: {data: T}) => renderString(get(data)),
```
GauBen, line 24: makes sense

src/components/DataTable/MIGRATION.md
```diff
@@ -0,0 +1,198 @@
...
+// Old API
+renderRow={(row, renderCells) => {
+    const release = row.original; // ← leaked internal structure
+    return <TableRow>{renderCells()}</TableRow>;
+}}
+```
+
+This created two problems:
+
+1. **Tight coupling to the table library.** If we ever swap the underlying table library, every consumer that used `row.original`, `row.getIsSelected()`, or any other internal method would break. The public API should be stable regardless of what runs underneath.
```
GauBen, line 17: 👍

GauBen (dismissed):
All good!

Issue #634, opened 2023-05-03
MOON-373 - Create z-index tokens


jahia-ci:
> Imported from Jira, on Thu Jan 30, 2025
> Issue: [MOON-373](https://jira.jahia.org/browse/MOON-373) in project: Moonstone
> Priority: Unspecified Type: Story
> Reporter: @ffffffelix (Félix Vallé)
> Assignee: @ffffffelix (Félix Vallé)
> Created: Wed May 3, 2023, last updated: Mon Jan 22, 2024
> Status: Open
> - Code Runners - Sprint 23.21 started on Tue May 23, 2023 completed on Tue Jun 6, 2023
> - Code Runners - Sprint 23.23 started on Tue Jun 6, 2023 completed on Tue Jun 20, 2023
> - Code Runners - Sprint 23.25 started on Tue Jun 20, 2023 completed on Tue Jul 4, 2023
> - Code Runners - Sprint 23.27 started on Tue Jul 4, 2023 completed on Tue Jul 18, 2023
>
>
> **Links**:
> - relates to: [BACKLOG-21079 - Issue when displaying file renaming modal](https://jira.jahia.org/browse/BACKLOG-21079) (Closed)
> - relates to: [DESIGN-1300 - Review z-index configuration throughout the content-editor module](https://jira.jahia.org/browse/DESIGN-1300)
>

To prevent inconsistent z-index values, we must create z-index tokens with a clear list of values available according to the layer.

An example of inspiration: 

https://mui.com/material-ui/customization/z-index/

jahia-ci:
> Comment by: @rknj (Michael De Checchi) on Thu May 4, 2023 _edited_


Discussed with @jahia-ci (Thomas Draier <X>) @ffffffelix (Félix Vallé) and @jahia-carp (Cyril Arp)

It'll be a 3-steps task:
1. This ticket to create the tokens
1. Share them and use them in all new features
1. Use them opportunistically in existing source code while fixing a bug

Open questions:
* Which kind of dependency will it add?
* Is it OK to push that in maintenance versions of the modules and Jahia?

jahia-ci:
> Comment by: @rknj (Michael De Checchi) on Mon Jun 26, 2023 _edited_


After release of Moonstone, @ffffffelix (Félix Vallé) will communicate about that in #team-product

jahia-ci:
> Comment by: @rknj (Michael De Checchi) on Mon Jun 26, 2023 _edited_


[...]

PR #1393, opened 2026-07-23
fix: tooltip overflow

ffffffelix:
## Description
Issue: When the Tooltip is displayed in an overflow context, the tooltip was truncated.
Solution: Use `FloatingPortal` to display the tooltip in the HTML body element instead of the parent element.

## Description
Issue: When the Tooltip is rendered in an overflow context like a DataTable cell, it is truncated.

Solution:
- Use `FloatingPortal` to render the tooltip in the `<body>` element instead of inline in the parent.
- Raise the tooltip `z-index` to `9100` so it sits above the highest current layer (`9000`). The `z-index` values will be improved with https://github.com/Jahia/moonstone/issues/634

## Tests

- [ ] Unit Tests
- [ ] Accessibility is OK

## Checklist

- [ ] Example in storybook
- [ ] Allows custom props

### DIFF
--- a/src/components/Tooltip/Tooltip.module.scss
+++ b/src/components/Tooltip/Tooltip.module.scss
@@ -1,7 +1,7 @@
 $bgColor: var(--moon-color-gray_dark);
 
 .moonstone-tooltip_label {
-    z-index: 1;
+    z-index: 9100;
 
     width: inherit;
     max-width: 450px;
@@ -14,7 +14,7 @@ $bgColor: var(--moon-color-gray_dark);
 }
 
 .moonstone-tooltip_arrow {
-    z-index: 1;
+    z-index: 9100;
 
     fill: $bgColor;
 }
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -68,8 +68,7 @@ export const Tooltip = ({
                                 {label}
                             </Typography>
                         </div>
-                    </FloatingPortal>
-                }
+                    </FloatingPortal>}
             </div>
         );
     }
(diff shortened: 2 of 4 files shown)


src/components/Tooltip/Tooltip.module.scss
```diff
@@ -1,7 +1,7 @@
 $bgColor: var(--moon-color-gray_dark);
 
 .moonstone-tooltip_label {
-    z-index: 1;
+    z-index: 9100;
```
BouklifYacine, line 4: Why do we use this big value for the index? 

Floating portal doesnt handle the problem itself?

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -14,7 +14,7 @@ $bgColor: var(--moon-color-gray_dark);
 }
 
 .moonstone-tooltip_arrow {
-    z-index: 1;
+    z-index: 9100;
```
BouklifYacine, line 17: same

BouklifYacine (changes requested):

src/components/Tooltip/Tooltip.tsx
```diff
@@ -54,19 +54,22 @@ export const Tooltip = ({
...
-                    <Typography>
-                        {label}
-                    </Typography>
-                </div>}
+                    <FloatingPortal>
+                        <div
+                            ref={refs.setFloating}
+                            id="moonstone-tooltip_label"
+                            className={clsx('moonstone-tooltip_label', styles['moonstone-tooltip_label'])}
+                            style={floatingStyles}
```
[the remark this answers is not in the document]
ffffffelix: We should have only one Tooltip at a time. It's acceptable.

src/components/Tooltip/Tooltip.module.scss
ffffffelix: It's in case other elements use Portal too; we will not guarantee that the Tooltip will be displayed on top of them. (example: modal component).

BouklifYacine (approved):

PR #1394, opened 2026-07-23
feat: add storybook mcp

BouklifYacine:
- Adds @storybook/addon-mcp + .mcp.json so components/stories are queryable over MCP.
- Enables experimentalReactComponentMeta — precomputes component metadata into the manifest (~20× faster manifest reads for the LLM).
- Adds a storybook Vitest project (Playwright/Chromium) so stories run as tests.
- Wires up addon-a11y in todo mode (a11y violations surfaced, non-blocking).
- Fixes coverage to target source files instead of specs, so the widget shows a real % instead of NaN%.


vite.config.mjs
```diff
@@ -6,13 +6,15 @@ import react from '@vitejs/plugin-react';
 import sbom from 'rollup-plugin-sbom';
 import {playwright} from '@vitest/browser-playwright';
 import {patchCssModules} from 'vite-css-modules';
+import {fileURLToPath} from 'node:url';
+import {storybookTest} from '@storybook/addon-vitest/vitest-plugin';
+const dirname = typeof __dirname !== 'undefined' ? __dirname : path.dirname(fileURLToPath(import.meta.url));
 
+// More info at: https://storybook.js.org/docs/next/writing-tests/integrations/vitest-addon
```
ffffffelix: We don't need that comment.

mcp-incompatible-components.md
```diff
@@ -0,0 +1,70 @@
+# Components not detected by the MCP
```
ffffffelix: I would prefer a comment in the PR itself or a specific ticket to handle it. Instead of the file directly into the repository.

vite.config.mjs
```diff
@@ -33,73 +35,105 @@ export default defineConfig({
...
     },
     assetsInclude: ['**/*.md'],
     test: {
         coverage: {
             provider: 'v8',
-            include: ['src/**/*.spec.tsx'],
-            exclude: ['src/__mocks__', 'src/__storybook__', 'src/data', '**/*.stories.*']
+            include: ['src/**/*.{ts,tsx}'],
+            exclude: ['src/__mocks__', 'src/__storybook__', 'src/data', '**/*.stories.*', '**/*.spec.tsx']
         },
```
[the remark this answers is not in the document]
ffffffelix: Thanks for fixing the coverage stat issue! 🙏
Agree with that comment.
Maybe we also need to exclude generated icons: `src/icons/components/*` ?

vite.config.mjs
```diff
@@ -6,13 +6,15 @@ import react from '@vitejs/plugin-react';
 import sbom from 'rollup-plugin-sbom';
 import {playwright} from '@vitest/browser-playwright';
 import {patchCssModules} from 'vite-css-modules';
+import {fileURLToPath} from 'node:url';
+import {storybookTest} from '@storybook/addon-vitest/vitest-plugin';
+const dirname = typeof __dirname !== 'undefined' ? __dirname : path.dirname(fileURLToPath(import.meta.url));
```
ffffffelix: You don't need that.

vite.config.mjs
```diff
@@ -33,73 +35,105 @@ export default defineConfig({
...
                     }
                 }
+            }
+        }, {
+            extends: true,
+            plugins: [
+                // The plugin will run tests for the stories defined in your Storybook config
+                // See options at: https://storybook.js.org/docs/next/writing-tests/integrations/vitest-addon#storybooktest
+                storybookTest({
+                    configDir: path.join(dirname, '.storybook')
```
ffffffelix:
```suggestion
                    configDir: path.resolve('.storybook')
```

vite.config.mjs
```diff
@@ -33,73 +35,105 @@ export default defineConfig({
...
-                            }
-                        }
-                    }
+            test: {
+                name: 'storybook',
+                setupFiles: ['./.storybook/vitest.setup.ts'],
+                browser: {
+                    enabled: true,
+                    headless: true,
+                    provider: playwright({}),
```
ffffffelix:
```suggestion
                    provider: playwright(),
```

vitest.shims.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="@vitest/browser-playwright" />
```
ffffffelix: I don't sure we really need it.

ffffffelix (changes requested):
Be careful with your IDE formatter, as the diff is hard to read when it reflows entire files.

.storybook/vitest.setup.ts
```diff
@@ -0,0 +1,9 @@
+import { beforeAll } from 'vitest';
+import { setProjectAnnotations } from '@storybook/react-vite';
+import * as projectAnnotations from './preview';
+
+// Applies Storybook's project-level annotations (decorators, globals, parameters
+// from preview.jsx) to every story rendered by the Vitest addon.
+const project = setProjectAnnotations([projectAnnotations]);
+
+beforeAll(project.beforeAll);
```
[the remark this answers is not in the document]
ffffffelix: I thought we needed it, no?

vite.config.mjs
```diff
@@ -99,6 +100,28 @@ export default defineConfig({
...
+                extends: true,
+                plugins: [
+                    // Runs the tests for the stories defined in your Storybook config
+                    storybookTest({configDir: path.resolve('.storybook')})
+                ],
+                // Pre-bundle CJS-only deps so their named exports resolve in browser mode
+                // (e.g. aria-query's `elementRoles`, pulled in transitively by the a11y setup).
+                optimizeDeps: {
+                    include: ['aria-query']
+                },
```
ffffffelix: Not sure if we need that.

ffffffelix (changes requested):

vite.config.mjs
BouklifYacine: I cant answer for vitest file on the com so i will do it here : 

Since Storybook 10.3 the addon applies annotations and cleanup
automatically, so the line would do nothing. Keeping `beforeAll` only.

BouklifYacine: For aria : Removed. Vite pre-bundles aria-query on its own, so the explicit include was
redundant — good catch. Side note: the story tests wouldn't boot on Windows (i dont know why), so I verified this on WSL/Linux where they
run fine.

ffffffelix (approved):

PR #1397, opened 2026-07-28
chore: use CSF-next format for stories

### LINKED ISSUE #1211: Write stories with CSF next format

BouklifYacine:
- Migrated 66+ stories to the CSF next format (`preview.meta()` factory pattern)
- Cleaned up now-unused type imports (types are now inferred automatically)
- Excluded `*.stories.tsx` from `tsconfig.json` to fix a `JavaScript heap out of memory` error on `yarn build:lib` ( this only affects `.d.ts` generation for the published lib, Storybook itself is unaffected since it has its own build config (`.storybook/main.ts`) )


.storybook/preview.jsx
```diff
@@ -1,83 +1 @@
...
-        // 'todo' - show a11y violations in the test UI only
-        // 'error' - fail CI on a11y violations
-        // 'off' - skip a11y checks entirely
-        test: "todo",
-    },
-};
-
-setupBackgroundListener();
-export const tags = ["autodocs"];
+export {default} from '../src/__storybook__/preview';
```
ffffffelix, line 1: I would prefer to keep the file here in `./storybook` and write it in TypeScript instead of changing the location.
If the goal is to simplify import, we can create a new alias `storybook` to point to `./storybook/*`

src/__storybook__/preview.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+function setupBackgroundListener() {
+    channel.removeListener(STORY_ARGS_UPDATED, storyListener);
+    channel.addListener(STORY_ARGS_UPDATED, storyListener);
+}
+
+setupBackgroundListener();
+
+const tagBadgesAddon = definePreviewAddon(storybookAddonTagBadges);
+const legacyDocsParametersAddon = definePreviewAddon<{
```
ffffffelix: Maybe we should rename `legacyDocsParametersAddon`, it's not legacy. It's the current config.
I suggest : `docsParametersAddon`

tsconfig.json
```diff
@@ -39,6 +39,7 @@
     "node_modules",
     "build",
     "dist",
-    "types"
+    "types",
+    "src/**/*.stories.tsx"
```
ffffffelix, line 43: I have the same OOM issue, but I don't like excluding `stories` because they don't typecheck anymore, so we lose the benefit of the CSF-next migration that should improve story typing.
Can you check whether we can avoid the OOM issue without excluding stories?

ffffffelix (changes requested):

tsconfig.json
BouklifYacine: Storybook is working on this solution they seem to have the same problem so we are waiting for a fix for now.

PR #1405, opened 2026-08-03
feat: make field description formattable

BouklifYacine:
- Helper prop (Field, FieldBoolean, Fieldset) from string to React.ReactNode type
- Add story + test per component

### DIFF
--- a/.chachalog/-tdMIuYw.md
+++ b/.chachalog/-tdMIuYw.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Allow ReactNode for Field/Fieldset/FieldBoolean helper prop (#1405)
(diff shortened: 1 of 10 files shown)


.chachalog/-tdMIuYw.md
```diff
@@ -0,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
```
ffffffelix: We can set it to `patch`

ffffffelix (commented):
It would be better if we could have just some words formatted, rather than the full description, to match real usage.

.chachalog/-tdMIuYw.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Allow ReactNode for Field/Fieldset/FieldBoolean helper prop (#1405)
```
ffffffelix:
```suggestion
Allow any ReactNode as `helper` for Field, Fieldset, and FieldBoolean to allow formatted description (#1405)
```

ffffffelix (approved):

PR #1410, opened 2026-08-06
feat: deprecate FieldSelector for field row CSS classes

BouklifYacine:
- Field rows are now built with CSS classes instead of the FieldSelector wrapper.
- Field, Fieldset, DynamicFieldset, Modal and CustomDropdown stories migrated, no visual change.
- FieldSelector is deprecated but still exported: no breaking change.


.chachalog/aDYpp8RR.md
```diff
@@ -3,4 +3,4 @@
 "@jahia/moonstone": minor
 ---
 
-Add `DateTimeInput`, `TimeInput`, and `TimezoneSelector` in beta (#1318)
+Deprecate FieldSelector for field row CSS classes (#1410)
```
ffffffelix:
```suggestion
Deprecate `FieldSelector` component (#1410)
```

src/components/Dropdown/CustomDropdown/CustomDropdown.stories.tsx
```diff
@@ -1,10 +1,12 @@
 import React, {useState} from 'react';
+import clsx from 'clsx';
 import {action} from 'storybook/actions';
 import {CustomDropdown} from './CustomDropdown';
 import type {CustomDropdownProps} from './CustomDropdown.types';
-import {Button, Fieldset, CardSelector, Chip, Dropdown, EmptyCardSelector, Field, FieldBoolean, FieldSelector, Input, MenuItem, RadioGroup, RadioItem, Separator, Textarea, Typography} from '~/components';
+import {Button, Fieldset, CardSelector, Chip, Dropdown, EmptyCardSelector, Field, FieldBoolean, Input, MenuItem, RadioGroup, RadioItem, Separator, Textarea, Typography} from '~/components';
 import * as icons from '~/icons/components';
 import {File} from '~/icons';
+import fieldStyles from '~/components/Field/Field.module.scss';
```
ffffffelix: When you write stories, you have the role of a user who consumes our library. 
So you must not use internal stuff in stories, except for the story's layout.

src/components/Dropdown/CustomDropdown/CustomDropdown.stories.tsx
```diff
@@ -95,45 +97,69 @@ export const Form = {
     <Fieldset id="form" label="Form">
         <Field id="field-multiple" label="Field" chips={<><Chip color="accent" label="Required"/><Chip icon={<icons.Language/>} label="Shared by all languages"/></>} buttons={<><Button icon={<icons.Add/>} label="Add"/><Button icon={<icons.MoreVert/>} variant="ghost"/></>} helper="information">
             <>
-                <FieldSelector buttons={<Button icon={<icons.MoreVert/>}/>} selector={<Input size="big" placeholder="Input value"/>}/>
-                <FieldSelector buttons={<Button icon={<icons.MoreVert/>}/>} selector={<Input size="big" placeholder="Input value"/>}/>
-                <FieldSelector buttons={<Button icon={<icons.MoreVert/>}/>} selector={<Input size="big" placeholder="Input value"/>}/>
+                <div className={clsx('moonstone-field_row', fieldStyles['moonstone-field_row'])}>
```
ffffffelix: Do you think it's the usage we want to recommend as best practices for using our component?

src/components/Dropdown/Dropdown.module.scss
```diff
@@ -4,7 +4,10 @@
 @use "../../utils/mixins" as *;
 
 .moonstone-dropdown_container {
-    composes: moonstone-dropdown_container from "../Field/FieldSelector/FieldSelector.module.scss";
+    // Remove the FieldSelector source with the deprecated component
+    composes:
+        moonstone-dropdown_container from "../Field/FieldSelector/FieldSelector.module.scss",
```
ffffffelix: You still use `FieldSelector.module.scss`. Did I miss something?

src/components/Field/FieldSelector/FieldSelector.md
```diff
@@ -1,3 +1,5 @@
+**Deprecated**: use the `moonstone-field_row` CSS classes documented on Field instead.
```
ffffffelix: We don't want to recommend using our class. We plan to be fully CSS modules. We don't want consumers to use our classes directly.

src/components/Field/FieldSelector/FieldSelector.stories.tsx
```diff
@@ -8,7 +8,7 @@ import {Close, MoreVert} from '~/icons';
 const meta: Meta<typeof FieldSelector> = {
     title: 'Components/Field/FieldSelector',
     component: FieldSelector,
-    tags: ['beta'],
+    tags: ['deprecated'],
```
ffffffelix, line 11: Good idea !

src/components/Field/FieldSelector/FieldSelector.tsx
```diff
@@ -5,13 +5,18 @@ import {HandleDrag} from '~/icons';
 import {layout} from '~/globals/css-utils.js';
 import styles from './FieldSelector.module.scss';
 
+/**
+ * @deprecated Use the `moonstone-field_row` CSS classes instead, they are documented on `Field`.
```
ffffffelix: No, we don't recommend using any moonstone classes. We recommend using the `children` prop of `Field` directly. But we don't recommend anything about how to structure it.

src/components/Field/FieldSelector/FieldSelector.types.ts
```diff
@@ -13,6 +16,8 @@ export type FieldSelectorProps = Omit<React.ComponentPropsWithRef<'div'>, 'class
 
     /**
      * FieldSelector isDraggable
+     *
+     * @deprecated Only displays the drag handle, it does not handle the drag and drop.
```
ffffffelix: As the component itself is deprecated, I don't think we need to deprecate the prop.

src/components/Field/Field.module.scss
```diff
@@ -35,3 +35,45 @@
         color: var(--moon-color-warning_dark);
     }
 }
+
```
ffffffelix: We don't need these classes.

ffffffelix (changes requested):
You may have misunderstood the task.
I don't want to replace `FieldSelector` with classes; it's no longer our responsibility. We already ship many tools to do it (flex classes ...).
I just want to simplify the usage of the `Field` component and pass the selector into the `children` prop as you did for the tests. In our side I just want to set the right style on `moonstone-field_children` to stack element in the right alignement and gap.

src/components/Dropdown/Dropdown.module.scss
BouklifYacine: Removing it would shrink it to 112px for apps still using the deprecated component, so it goes away with the component.

PR #1415, opened 2026-08-12
fix: better position for the datetimeinput calendar

### LINKED ISSUE #1414: Improve the position of the datetimeinput calendar
The way the calendar of datetimeinput component opens is not ideal, part of it may be cut off depending on the screen or page.

Suggestions:
- Check props of Menu to fix it.
- Don't use Menu and use Floating-ui (already loaded in moonstone)
- Maybe another solution

BouklifYacine:
- Use of floating ui to handle the position of the datetimeinput calendar

### DIFF
--- a/src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
+++ b/src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
@@ -1,9 +1,10 @@
-import React, {useCallback, useState} from 'react';
+import React, {useState} from 'react';
 import clsx from 'clsx';
+import {autoUpdate, flip, FloatingPortal, offset, shift, useDismiss, useFloating, useInteractions, useMergeRefs} from '@floating-ui/react';
 import {dateMatchModifiers, DayPicker} from 'react-day-picker';
 import dayPickerClassNames from 'react-day-picker/style.module.css';
 import {Temporal} from 'temporal-polyfill';
-import {Button, Dropdown, Menu, Typography} from '~/components';
+import {Button, Dropdown, Typography} from '~/components';
 import {Calendar} from '~/icons';
 import type {DropdownProps} from 'react-day-picker';
 import {TimezoneSelector} from '../../TimezoneSelector/TimezoneSelector';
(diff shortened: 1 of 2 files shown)


src/components/Input/DateTimeInput/DateTimeInput.module.scss
```diff
@@ -12,6 +10,18 @@
...
 }
 
+.calendarPopover {
+    z-index: 9000;
+
+    width: max-content;
+    max-height: calc(100vh - 16px);
+    overflow-y: auto;
+
+    border-radius: var(--moon-radius);
```
ffffffelix: Do we really need these styles?

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
```diff
@@ -79,15 +80,16 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
...
-        } else if (ref) {
-            (ref as React.MutableRefObject<HTMLInputElement | null>).current = node;
-        }
-    }, [ref]);
+    const {refs, floatingStyles, context} = useFloating({
+        open: isCalendarOpen,
+        onOpenChange: setIsCalendarOpen,
+        placement: 'bottom-start',
+        transform: false,
+        middleware: [offset(4), flip({padding: 8}), shift({padding: 8})],
```
ffffffelix: - The calendar is not aligned with the input
- In some cases, the calendar is still truncated and doesn't "flip" above the input when there is not enough space below.

ffffffelix (changes requested):

src/components/Input/DateTimeInput/DateTimeInput.module.scss
BouklifYacine: Overflow is needed because without this the calendar go above the input

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
```diff
@@ -33,6 +35,10 @@ import {
 import type {ControlledDateTimeInputProps} from './DateTimeInput.types';
 import styles from './DateTimeInput.module.scss';
 
+const capCalendarHeight: SizeOptions['apply'] = ({availableHeight, elements}) => {
```
ffffffelix: I don't get whu you need to calculate / specify the height of the calendar ?

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
```diff
@@ -133,6 +139,7 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
 
     const openCalendar = () => {
         if (!isDisabled && !isReadOnly) {
+            refs.setPositionReference(refs.domReference.current?.closest('.moonstone-baseInput') ?? null);
```
ffffffelix: Don't use CSS classes as a selector, at leaset use a css module classe.

src/components/Input/DateTimeInput/DateTimeInput.module.scss
```diff
@@ -12,6 +10,13 @@
     width: 140px;
 }
 
+.calendarPopover {
+    overflow-y: auto;
```
ffffffelix: I don't think we should allow scrolling into this `div`, but in the calendar itself, even if I would like to avoid adding a scrollbar for the calendar.

ffffffelix (changes requested):

src/components/Input/DateTimeInput/DateTimeInput.module.scss
ffffffelix: I just saw your previous comment:
>Overflow is needed because without this the calendar go above the 

But I don't get why adding overflow changes the calendar position 🤔

BouklifYacine: It's was a bug that i have discovered during the implementation in jcontent

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
BouklifYacine: Without it, the calendar can overflow the screen near the edges, capCalendarHeight caps its height so it can scroll internally.

src/components/Input/DateTimeInput/DateTimeInput.module.scss
ffffffelix: Okay,
> But I don't get why adding overflow changes the calendar position 🤔

So I am not sure that was the good fix for the issue you mention. We will have to investigate later.

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
ffffffelix: We shouldn't avoid having to scroll **inside** the calendar

ffffffelix (dismissed):

ffffffelix (approved):

PR #1416, opened 2026-08-12
feat: Add the possibility to type a date in datetimeinput

BouklifYacine:
- The date can now be typed in the input, not only picked in the calendar.

### DIFF
--- a/.chachalog/Dt9mKqvB.md
+++ b/.chachalog/Dt9mKqvB.md
@@ -1,6 +1,0 @@
----
-# Allowed version bumps: patch, minor, major
-"@jahia/moonstone": minor
----
-
-Allow typing a date manually in the `DateTimeInput` field
--- a/.chachalog/Dt9mKqvB.md
+++ b/.chachalog/Dt9mKqvB.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Allow typing a date manually in the `DateTimeInput` field
--- a/.chachalog/pQYSDc6d.md
+++ b/.chachalog/pQYSDc6d.md
@@ -1,6 +1,0 @@
----
-# Allowed version bumps: patch, minor, major
-"@jahia/moonstone": minor
----
-
-Add the possibility to type a date in datetimeinput (#1416)
--- a/.chachalog/pQYSDc6d.md
+++ b/.chachalog/pQYSDc6d.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Add the possibility to type a date in datetimeinput (#1416)
--- a/src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
+++ b/src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
@@ -65,6 +65,7 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
     timeInputProps,
     timezoneSelectorProps,
     onBlur,
+    autoComplete = 'off',
     ...props
 }, ref) => {
     const currentValue = parseValue(value, type);
@@ -195,7 +196,7 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
                 variant={variant}
                 isDisabled={isDisabled}
                 isReadOnly={isReadOnly}
-                autoComplete="off"
+                autoComplete={autoComplete}
                 icon={<Calendar aria-hidden/>}
                 onChange={event => setDraft(event.target.value)}
                 onClear={event => {
--- a/src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
+++ b/src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
@@ -19,9 +19,11 @@ import {
 import {
     formatPlainDate,
     getCalendarDisabledMatchers,
+    getDateFormat,
     getDisplayMonth,
     getMonthStart,
-    getWeekStartsOn
+    getWeekStartsOn,
+    parseDateInput
 } from './calendarHelpers';
 import {
     assembleValue,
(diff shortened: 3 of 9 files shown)


.chachalog/Dt9mKqvB.md
```diff
@@ -0,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/moonstone": minor
+---
+
+Allow typing a date manually in the `DateTimeInput` field
```
ffffffelix: As the component is not released yet, we don't need a changelog.

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
```diff
@@ -72,7 +75,10 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
         ...i18n
     };
 
+    const format = getDateFormat(resolvedLocale, dateFormat);
+
     const [isCalendarOpen, setIsCalendarOpen] = useState(false);
+    const [draft, setDraft] = useState<string | null>(null);
```
ffffffelix: You cannot use a controlled component with a state.

src/components/Input/DateTimeInput/calendarHelpers.ts
```diff
@@ -91,6 +91,34 @@ export const formatPlainDate = (value: Temporal.PlainDate | null, locale?: strin
     return new Intl.DateTimeFormat(locale || undefined).format(plainDateToDate(value));
 };
 
+const FIELD_RE = /YYYY|YY|MM|DD/g;
+
+const REFERENCE_DATE = Temporal.PlainDate.from('2033-11-22');
+
+export const getDateFormat = (locale: string, dateFormat?: DateFormat): string | undefined => {
```
ffffffelix: Please re-use what we already have, we shouldn't need it.

src/components/Input/DateTimeInput/calendarHelpers.ts
```diff
@@ -91,6 +91,34 @@ export const formatPlainDate = (value: Temporal.PlainDate | null, locale?: strin
     return new Intl.DateTimeFormat(locale || undefined).format(plainDateToDate(value));
 };
 
+const FIELD_RE = /YYYY|YY|MM|DD/g;
+
+const REFERENCE_DATE = Temporal.PlainDate.from('2033-11-22');
```
ffffffelix: Why ?

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
```diff
@@ -151,24 +182,26 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
         <div className={clsx(styles.dateTimeInput, className)}>
             <BaseInput
                 ref={handleRef}
+                placeholder={format}
                 {...props}
                 className={styles.dateField}
-                value={formatPlainDate(selectedDate, resolvedLocale, dateFormat)}
+                value={draft ?? formatPlainDate(selectedDate, resolvedLocale, dateFormat)}
```
ffffffelix: Not sure we always want to display `draft` as `value`

ffffffelix (changes requested):
- The date input should be committed when the user presses `Enter`
- We should have the same behavior as the time input.
- DateTimeInput.md isn't updated. It still describes the field as "a text trigger that opens a calendar", no mention of typing.
- Before adding function check if we already have something similar to use.

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
ffffffelix: Still don't get why you want to add the `draft` as `value`

src/components/Input/DateTimeInput/ControlledDateTimeInput.tsx
```diff
@@ -153,25 +181,31 @@ export const ControlledDateTimeInput = React.forwardRef<HTMLInputElement, Contro
                 ref={handleRef}
                 {...props}
                 className={styles.dateField}
-                value={formatPlainDate(selectedDate, resolvedLocale, dateFormat)}
+                value={draft ?? formatPlainDate(selectedDate, resolvedLocale, dateFormat)}
                 size={size}
                 variant={variant}
                 isDisabled={isDisabled}
                 isReadOnly={isReadOnly}
+                autoComplete="off"
```
ffffffelix: I think we can `autoComplete="off"` but it should be overridable throw the `props`

src/components/Input/DateTimeInput/DateTimeInput.spec.tsx
```diff
@@ -423,19 +423,13 @@ describe('DateTimeInput', () => {
         expect(dateField()).toHaveValue('');
     });
 
-    it('should open the calendar via keyboard (Enter and Space)', async () => {
+    it('should open the calendar via keyboard (Enter on an untouched field)', async () => {
```
ffffffelix: Why did we lose the `Space` to display the calendar ? ([reference](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/))

.chachalog/pQYSDc6d.md
```diff
@@ -3,4 +3,4 @@
 "@jahia/moonstone": minor
 ---
 
-Allow typing a date manually in the `DateTimeInput` field
\ No newline at end of file
+Add the possibility to type a date in datetimeinput (#1416)
```
ffffffelix: We don't need a changelog because the component hasn't been released yet.

src/components/Input/DateTimeInput/DateTimeInput.types.ts
```diff
@@ -133,6 +133,12 @@ export type DateTimeInputSharedProps = Omit<BaseInputProps,
      */
     weekStartsOn?: DayOfWeek;
 
+    /**
+     * Displays React DayPicker's built-in month dropdown in the calendar header.
+     * The month remains plain text when omitted.
+     */
+    isShowMonthDropdown?: boolean;
```
ffffffelix: It's a good idea to add the month dropdown, but I don't think we need to add a toggle to show or hide it. 
As a design system, we can always display the month selector when several months are allowed.
So you can remove this prop.

ffffffelix (changes requested):
- Why you removed the `placeholder={format}` it was a good idea

ffffffelix (dismissed):

ffffffelix (approved):

PR #1417, opened 2026-08-13
feat: allow searching in the DataTable

BouklifYacine:
- The Datatable can now display a search field that filters its rows as the user types, ignoring case.
- You choose which columns the search looks into.
- The search can also be handed off to the server: pass a controlled searchValue/onSearchChange (and skip searchColumns) if you filter the data yourself.
- A search returning no result no longer hides the search field, so it can always be cleared.

### DIFF
--- a/src/components/DataTable/DataTable.stories.tsx
+++ b/src/components/DataTable/DataTable.stories.tsx
@@ -18,7 +18,9 @@ export default {
         enablePagination: {control: 'boolean'},
         defaultItemsPerPage: {control: 'number'},
         itemsPerPageOptions: {control: 'object'},
-        i18n: {control: 'object'}
+        i18n: {control: 'object'},
+        enableSearch: {control: 'boolean'},
+        searchColumns: {control: 'object'}
     }
 } satisfies Meta<typeof DataTable<DataUser>>;
 
@@ -114,6 +116,75 @@ export const ControlledDataTable: Story = {
         );
     },
     name: 'Controlled DataTable'
+};
+
+export const SearchableDataTable: Story = {
+    render: args => {
+        return <DataTable {...args}/>;
+    },
+    args: {
+        data: tableFlat,
+        columns: dataColumnsUser,
+        primaryKey: 'id',
+        enableSearch: true,
+        searchColumns: ['firstName', 'status', 'progress'],
+        searchInputProps: {placeholder: 'Search by user, status or progress'}
+    },
+    name: 'Searchable DataTable (uncontrolled)'
+};
+
+export const ControlledSearchDataTable: Story = {
+    render: () => {
+        const [searchValue, setSearchValue] = useState('');
+
+        return (
+            <DataTable
+                enableSearch
+                data={tableFlat}
+                columns={dataColumnsUser}
+                primaryKey="id"
+                searchColumns={['firstName', 'status', 'progress']}
+                searchValue={searchValue}
+                onSearchChange={setSearchValue}
+            />
+        );
+    },
+    name: 'Controlled search (the table filters)'
+};
+
+export const ServerSideSearchDataTable: Story = {
+    render: () => {
+        const [searchValue, setSearchValue] = useState('');
+        const [currentPage, setCurrentPage] = useState(1);
+        const [itemsPerPage, setItemsPerPage] = useState(10);
+
+        const query = searchValue.toLowerCase();
+        const matchingRows = tableFlat.filter(row => `${row.firstName} ${row.lastName}`.toLowerCase().includes(query));
+        const pageRows = matchingRows.slice((currentPage - 1) * itemsPerPage, currentPage * itemsPerPage);
+
+        return (
+            <DataTable
+                enableSearch
+                enablePagination
+                enableSorting={false}
+                data={pageRows}
+                columns={dataColumnsUser}
+                primaryKey="id"
+                searchValue={searchValue}
+                currentPage={currentPage}
+                itemsPerPage={itemsPerPage}
+                totalItems={matchingRows.length}
+                searchInputProps={{placeholder: 'Search on the server'}}
+                onSearchChange={value => {
+                    setSearchValue(value);
+                    setCurrentPage(1);
+                }}
+                onPageChange={setCurrentPage}
+                onItemsPerPageChange={setItemsPerPage}
+            />
+        );
+    },
+    name: 'Server-side search (the consumer filters)'
 };
 
 export const InsertCells: Story = {
--- a/src/components/DataTable/DataTable.stories.tsx
+++ b/src/components/DataTable/DataTable.stories.tsx
@@ -133,25 +133,6 @@ export const SearchableDataTable: Story = {
     name: 'Searchable DataTable (uncontrolled)'
 };
 
-export const ControlledSearchDataTable: Story = {
-    render: () => {
-        const [searchValue, setSearchValue] = useState('');
-
-        return (
-            <DataTable
-                enableSearch
-                data={tableFlat}
-                columns={dataColumnsUser}
-                primaryKey="id"
-                searchColumns={['firstName', 'status', 'progress']}
-                searchValue={searchValue}
-                onSearchChange={setSearchValue}
-            />
-        );
-    },
-    name: 'Controlled search (the table filters)'
-};
-
 export const ServerSideSearchDataTable: Story = {
     render: () => {
         const [searchValue, setSearchValue] = useState('');
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -2,9 +2,11 @@ import {
     useReactTable,
     getCoreRowModel,
     getExpandedRowModel,
+    getFilteredRowModel,
     getSortedRowModel,
     getPaginationRowModel
 } from '@tanstack/react-table';
+import clsx from 'clsx';
 import {toNodeArray} from '~/utils/helpers';
 import type {Row} from '@tanstack/react-table';
 import React, {useMemo, useCallback} from 'react';
@@ -13,7 +15,9 @@ import type {DataTableProps, RenderOptions} from './DataTable.types';
 import {createTableColumns} from './shared';
 import {renderCell, renderHeadCell} from './utils';
 import {Checkbox} from '~/components';
+import {SearchInput} from '~/components/Input';
 import {Pagination} from '~/components/Pagination';
+import styles from './DataTable.module.scss';
 import {
     Table,
     TableRow,
@@ -60,6 +64,11 @@ export const DataTable = <T extends NonNullable<unknown>>({
     totalItems,
     i18n,
     paginationProps,
+    enableSearch = false,
+    searchColumns,
+    searchValue,
+    onSearchChange,
+    searchInputProps,
     rowProps,
     ...props
 }: DataTableProps<T>) => {
@@ -114,7 +123,8 @@ export const DataTable = <T extends NonNullable<unknown>>({
             expanded,
             rowSelection,
             sorting,
-            ...(enablePagination && {pagination})
+            ...(enablePagination && {pagination}),
+            ...(searchValue !== undefined && {globalFilter: searchValue})
         },
         onSortingChange: handleSortingChange,
         onExpandedChange: handleExpandedChange,
@@ -122,6 +132,10 @@ export const DataTable = <T extends NonNullable<unknown>>({
         getCoreRowModel: getCoreRowModel(),
         getSortedRowModel: enableSorting ? getSortedRowModel() : undefined,
         manualSorting: isSortingControlled,
+        getFilteredRowModel: enableSearch ? getFilteredRowModel() : undefined,
+        globalFilterFn: 'includesString',
+        getColumnCanGlobalFilter: column => Boolean(searchColumns?.some(key => key === column.id)),
+        filterFromLeafRows: isStructured,
         getExpandedRowModel: getExpandedRowModel(),
         getPaginationRowModel: enablePagination ? getPaginationRowModel() : undefined,
         manualPagination: isPaginationControlled,
@@ -202,12 +216,27 @@ export const DataTable = <T extends NonNullable<unknown>>({
         [renderRow, renderRowContent, rowProps]
     );
 
-    if (!data || !Array.isArray(data) || data.length === 0) {
+    const isEmpty = !data || !Array.isArray(data) || data.length === 0;
+
+    const searchQuery = String(table.getState().globalFilter ?? '');
+    const setSearchQuery: (value: string) => void = onSearchChange ?? table.setGlobalFilter;
+
+    if (isEmpty && !enableSearch) {
         return null;
     }
 
     return (
         <>
+            {enableSearch && (
+                <SearchInput
+                    variant="outlined"
+                    value={searchQuery}
+                    onChange={event => setSearchQuery(event.target.value)}
+                    onClear={() => setSearchQuery('')}
+                    {...searchInputProps}
+                    className={clsx(styles.search, searchInputProps?.className)}
+                />
+            )}
             <Table className={className} {...props}>
                 <TableHead>
                     {table.getHeaderGroups().map(headerGroup => (
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -119,6 +119,7 @@ export const DataTable = <T extends NonNullable<unknown>>({
     const table = useReactTable({
         data,
         columns: tableColumns,
+        initialState: {globalFilter: ''},
         state: {
             expanded,
             rowSelection,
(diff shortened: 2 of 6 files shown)


src/components/DataTable/DataTable.tsx
```diff
@@ -202,12 +217,30 @@ export const DataTable = <T extends NonNullable<unknown>>({
...
     );
 
-    if (!data || !Array.isArray(data) || data.length === 0) {
+    const isEmpty = !data || !Array.isArray(data) || data.length === 0;
+
+    const searchQuery = String(table.getState().globalFilter);
+    const setSearchQuery: (value: string) => void = onSearchChange ?? table.setGlobalFilter;
+    const searchLabel = searchInputProps?.placeholder ?? 'Search';
+
+    if (isEmpty && !enableSearch) {
```
ffffffelix: - Maybe it should be better to change the condition and remove `data.length === 0; otherwise, what will we display in the case of `enableSearch: true` and no `data` ?
- You don't display something when search match nothing ? we only have table header/footer. We should display something !

src/components/DataTable/DataTable.tsx
```diff
@@ -202,12 +217,30 @@ export const DataTable = <T extends NonNullable<unknown>>({
         [renderRow, renderRowContent, rowProps]
     );
 
-    if (!data || !Array.isArray(data) || data.length === 0) {
+    const isEmpty = !data || !Array.isArray(data) || data.length === 0;
+
+    const searchQuery = String(table.getState().globalFilter);
+    const setSearchQuery: (value: string) => void = onSearchChange ?? table.setGlobalFilter;
+    const searchLabel = searchInputProps?.placeholder ?? 'Search';
```
ffffffelix: So the prop should be named `searchPlaceholder` ?

src/components/DataTable/DataTable.tsx
```diff
@@ -202,12 +217,30 @@ export const DataTable = <T extends NonNullable<unknown>>({
...
         return null;
     }
 
     return (
         <>
+            {enableSearch && (
+                <SearchInput
+                    variant="outlined"
+                    value={searchQuery}
+                    aria-label={searchLabel}
```
ffffffelix: You cannot use the placeholder as `aria-label,` just let this responsibility to our user.

src/components/DataTable/DataTable.tsx
```diff
@@ -202,12 +217,30 @@ export const DataTable = <T extends NonNullable<unknown>>({
         [renderRow, renderRowContent, rowProps]
     );
 
-    if (!data || !Array.isArray(data) || data.length === 0) {
+    const isEmpty = !data || !Array.isArray(data) || data.length === 0;
+
+    const searchQuery = String(table.getState().globalFilter);
+    const setSearchQuery: (value: string) => void = onSearchChange ?? table.setGlobalFilter;
```
ffffffelix: Why don't you use a state?

src/components/DataTable/DataTable.types.ts
```diff
@@ -181,6 +182,41 @@ export type SelectionProps =
...
+      }
+    | {
+          /** Enable search for the table */
+          enableSearch: true;
+          /** Columns to search (case-insensitive substring match on the raw value). Omit for server-side search. */
+          searchColumns: SearchColumns<T>;
+          searchValue?: never;
+          onSearchChange?: never;
+          /** Custom attributes added to the SearchInput element (pass `placeholder` to localize it) */
+          searchInputProps?: SearchInputAttributes;
```
ffffffelix: I don't get this part? Can you explain please

ffffffelix (changes requested):
Don't forget to link Pr to the ticket please.

ffffffelix (commented):
We also have to be careful to avoid the case when we mismatch controlled/uncontrolled pagination/search

src/components/DataTable/DataTable.tsx
BouklifYacine: Kept the early return: removing it changes every existing DataTable, which has always rendered nothing without data. But you were right that nothing was shown — aedfa3f0 adds a row spanning all columns with noResultsMessage; a general EmptyData empty state needs a design call, so separate ticket.

src/components/DataTable/DataTable.tsx
BouklifYacine: That line is gone: the placeholder is now a plain default (placeholder="Search") overridden through searchInputProps, which follows the convention of paginationProps/selectionCellProps and covers aria-label and id too, not just the placeholder. A dedicated searchPlaceholder would add a second way to set the same thing

src/components/DataTable/DataTable.types.ts
```diff
@@ -181,6 +182,73 @@ export type SelectionProps =
...
+ * `searchColumns` stays optional here, because a controlled query usually
+ * means you filter `data` yourself (a server-side search) and the table must
+ * not filter it a second time. Pass it anyway if you only want to drive the
+ * value from the outside, for instance to keep it in a URL, and still let the
+ * table do the filtering.
+ *
+ * Uncontrolled: the table owns the query, so `searchColumns` becomes
+ * required — nothing else would filter the rows, and the field would look
+ * broken. `onSearchChange` is then a plain observer.
+ */
```
ffffffelix: For me, it's not because the input is controlled that the data table is done server-side.

src/components/DataTable/DataTable.types.ts
```diff
@@ -181,6 +182,73 @@ export type SelectionProps =
...
 
+type SearchColumns<T extends NonNullable<unknown>> = Array<Extract<Exclude<keyof T, SubRowKey>, string>>;
+
+type SearchInputAttributes = Omit<SearchInputProps, 'value' | 'defaultValue' | 'onChange' | 'onClear'>;
+
+/**
+ * Search accepts two shapes, and only two.
+ *
+ * Controlled: you own the query through `searchValue` and `onSearchChange`.
+ * `searchColumns` stays optional here, because a controlled query usually
```
ffffffelix: We should avoid "usually" in documentation. We need to avoid a blurry line.

src/components/DataTable/DataTable.tsx
```diff
@@ -26,6 +30,21 @@ import {
...
+    enablePagination: boolean;
+    isPaginationControlled: boolean;
+};
+
+// The types cannot express a constraint spanning two features, so it is checked at runtime
+const warnOnSearchPaginationMismatch = ({enableSearch, isSearchControlled, searchColumns, enablePagination, isPaginationControlled}: SearchPaginationGuard) => {
+    if (enableSearch && isSearchControlled && !searchColumns?.length && enablePagination && !isPaginationControlled) {
+        console.warn('A controlled `searchValue` without `searchColumns` means the DataTable component filters nothing: pagination must be controlled too (`currentPage` and `totalItems`), otherwise the item count only reflects the rows of the current page.');
+    }
+};
```
ffffffelix: Checks isSearchControlled (who holds the string) but the danger is who filters. Missing the mirror case: table filters + controlled pagination → "three rows, totalItems={300}, typing alice renders one row while the footer says 1-5 of 300, and nothing warns."

Also worth naming enablePagination={false} in the message, since it defaults to true.

src/components/DataTable/DataTable.tsx
```diff
@@ -202,12 +247,28 @@ export const DataTable = <T extends NonNullable<unknown>>({
         [renderRow, renderRowContent, rowProps]
     );
 
-    if (!data || !Array.isArray(data) || data.length === 0) {
+    const isEmpty = !data || !Array.isArray(data) || data.length === 0;
```
ffffffelix: With `enableSearch`, empty data and no query, we render No results + 1-0 of 0. "No results" claims a search happened. Gate it on a non-empty query.

src/components/DataTable/DataTable.tsx
ffffffelix: If it's a separate ticket, create it.

src/components/DataTable/DataTable.tsx
```diff
@@ -252,7 +313,13 @@ export const DataTable = <T extends NonNullable<unknown>>({
                     ))}
                 </TableHead>
                 <TableBody>
-                    {table.getRowModel().rows.map(row => renderRowWithCustomization(row))}
+                    {enableSearch && rows.length === 0 ? (
+                        <TableRow>
+                            <TableCell colSpan={columnCount} align="center">{noResultsMessage}</TableCell>
```
ffffffelix: Needs `aria-live="polite"`. A sighted user sees rows vanish while typing; a screen-reader user hears nothing.

src/components/DataTable/hooks/useSearch.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type UseSearchProps = {
+    searchValue?: string;
+    onSearchChange?: (searchValue: string) => void;
+};
+
+export function useSearch({searchValue, onSearchChange}: UseSearchProps) {
+    const isSearchControlled = searchValue !== undefined;
+
+    const [state, setState] = useState('');
```
ffffffelix: Add `defaultSearchValue` so a query can be seeded without going fully controlled.

src/components/DataTable/DataTable.types.ts
```diff
@@ -181,6 +182,73 @@ export type SelectionProps =
...
+          /**
+           * Custom attributes added to the SearchInput element.
+           * Pass `placeholder` to localize it, and `aria-label` or `aria-labelledby` to give it an accessible name.
+           */
+          searchInputProps?: SearchInputAttributes;
+          /**
+           * Message displayed instead of the rows when the query matches nothing
+           * @default 'No results'
+           */
+          noResultsMessage?: string;
```
ffffffelix: I am wondering if this prop should be into a `i18n` prop. The original idea was to handle translation labels from this prop.
It will be great to use the prop `i18n` at the `DataTable` level to have one entry to manage all translation related to the table.

src/components/DataTable/DataTable.tsx
```diff
@@ -202,12 +247,28 @@ export const DataTable = <T extends NonNullable<unknown>>({
...
         return null;
     }
 
     return (
         <>
+            {enableSearch && (
+                <SearchInput
+                    variant="outlined"
+                    value={globalFilter}
+                    placeholder="Search"
```
ffffffelix: Don't need to set a default placeholder here. Let's the product doing it.

src/components/DataTable/DataTable.types.ts
```diff
@@ -181,6 +182,73 @@ export type SelectionProps =
           selectionCellProps?: never;
       };
 
+type SearchColumns<T extends NonNullable<unknown>> = Array<Extract<Exclude<keyof T, SubRowKey>, string>>;
+
+type SearchInputAttributes = Omit<SearchInputProps, 'value' | 'defaultValue' | 'onChange' | 'onClear'>;
```
ffffffelix: In order to improve accessibility maybe we can require some props.
```ts
type SearchInputAttributes =
    Omit<SearchInputProps, 'value' | 'defaultValue' | 'onChange' | 'onClear'>
    & ({'aria-label': string} | {'aria-labelledby': string});
```

In that case `searchInputProps` has to become required when `enableSearch` is `true`, otherwise the object can simply be omitted and we are back to an input with no name.

src/components/DataTable/hooks/useSearch.ts
```diff
@@ -0,0 +1,26 @@
...
+    const globalFilter = isSearchControlled ? searchValue : state;
+
+    const handleGlobalFilterChange = (updater: React.SetStateAction<string>) => {
+        const next = typeof updater === 'function' ? updater(globalFilter) : updater;
+
+        if (!isSearchControlled) {
+            setState(next);
+        }
+
+        onSearchChange?.(next);
```
ffffffelix: `onSearchChange` is called even when the value did not change. For example, clicking "Reset" on a field that is already empty calls it with ''. `usePagination` checks first `(if (next.pageIndex !== pagination.pageIndex))`. Can we do the same here?

src/components/DataTable/DataTable.stories.tsx
```diff
@@ -116,6 +118,56 @@ export const ControlledDataTable: Story = {
...
+        columns: dataColumnsUser,
+        primaryKey: 'id',
+        enableSearch: true,
+        searchColumns: ['firstName', 'status', 'progress'],
+        searchInputProps: {placeholder: 'Search by user, status or progress'}
+    },
+    name: 'Searchable DataTable (uncontrolled)'
+};
+
+export const ServerSideSearchDataTable: Story = {
```
ffffffelix: This story does not use args, so its Storybook controls do nothing. Also, why enableSorting={false}? A short comment would help the reader.

src/components/DataTable/DataTable.stories.tsx
```diff
@@ -18,7 +18,9 @@ export default {
         enablePagination: {control: 'boolean'},
         defaultItemsPerPage: {control: 'number'},
         itemsPerPageOptions: {control: 'object'},
-        i18n: {control: 'object'}
+        i18n: {control: 'object'},
+        enableSearch: {control: 'boolean'},
```
ffffffelix: This control lets a reader turn enableSearch on for a story that has no searchColumns. The search box is then dead — see my comment on DataTable.types.ts. Maybe remove this control, or add searchColumns to those stories.

src/components/DataTable/DataTable.types.ts
```diff
@@ -181,6 +182,73 @@ export type SelectionProps =
...
+ * Uncontrolled: the table owns the query, so `searchColumns` becomes
+ * required — nothing else would filter the rows, and the field would look
+ * broken. `onSearchChange` is then a plain observer.
+ */
+type SearchProps<T extends NonNullable<unknown>> =
+    | {
+          /** Enable search for the table */
+          enableSearch: true;
+          /** Columns filtered by the table. Omit it when you filter `data` yourself */
+          searchColumns?: SearchColumns<T>;
```
ffffffelix: `SearchColumns<T>` is an Array (line 185). So `searchColumns={[]}` is valid TypeScript.

But an empty array means "no column can be searched". Two things then happen:
- `getColumnCanGlobalFilter` returns false for every column
- `manualFiltering` becomes true, so the table does not filter at all

I tested it. A table with 3 rows and `searchColumns={[]}`. I typed `zzzz`. The 3 rows stayed visible. No warning in the console.

So the user types in a search box and nothing happens, ever. Can we make this impossible in the type?
```
searchColumns: [SearchColumns<T>[number], ...SearchColumns<T>[number][]];
```
This asks for at least one column.

ffffffelix (changes requested):
### Controlled/uncontrolled and client/server are two different questions

1. **Who remembers the typed text?** The table (uncontrolled), or you (controlled: `searchValue` + `onSearchChange`). This is a plain React question about an input.
2. **Who removes the rows that do not match?** The table (it needs `searchColumns`), or you / your server (you pass `data` already filtered).

The PR answers question 2 by looking at question 1: in the controlled shape, the *absence* of `searchColumns` means "the table must not filter".

To be fair, that absence is honest — `searchColumns` really is meaningless when the server filters, because the field list lives in your own query. The problem is only that **"absent because meaningless" and "absent because forgotten" look exactly the same.**

Four combinations exist. Only three can be written:

| | The table filters | You filter |
|---|---|---|
| **The table remembers the text** | ✅ `enableSearch searchColumns` | ❌ cannot be written |
| **You remember the text** | ✅ `searchValue searchColumns` | ✅ `searchValue` |

The missing one is the shape a debounced server search wants: let the table own the box so typing feels instant, while you own the filtering.

### Two bugs come from this

**The row count lies.** Uncontrolled search (the table filters) + controlled pagination (you count). Both are legal, and nothing warns. Three rows, `totalItems={300}`, typed `alice`:

```
Name  Age
Alice  30
1-5 of 300      ← console.warn was called 0 times
```

The table filtered the page it holds. The footer still promises 300 items.

**The search box does nothing.** `searchColumns={[]}` compiles. An empty array means no column can be searched, so `manualFiltering` becomes `true`. Typed `zzzz` into 3 rows → all 3 still visible, `1-3 of 3`, no warning.

### What this PR needs
- make `searchColumns` a non-empty array, so the dead search box cannot happen
- add the missing case to the guard: the table filters **and** you paginate
- require `aria-label` or `aria-labelledby`, since the default `placeholder` is going away
- show "No results" only when there is actually a query

src/components/DataTable/DataTable.tsx
BouklifYacine: There's the ticket : https://github.com/Jahia/moonstone/issues/1460

src/components/DataTable/DataTable.types.ts
BouklifYacine: We finish what is in progress. The noResultsMessage prop is removed and "No results" is hardcoded for now. 
It will be resolve in this ticket : https://github.com/Jahia/moonstone/issues/1459

PR #1418, opened 2026-08-13
fix: improve accessibility of field component

BouklifYacine:
- Improve the accessibility for the field component ( aria-labelledby - aria-describedby )


ffffffelix (changes requested):
What about when the `id` is not provided ?

PR #1420, opened 2026-08-26
Demo branch to test with all fix and updates in jcontent

BouklifYacine:
This PR is only to be able to have the pkg link with all the last updates and fix to datetime component to test it in jcontent


BouklifYacine:
I close this PR it was only to test our changes of datetimeinput in Jcontent with the pkg link no needed anymore

